PR #156, opened 2024-02-23
Add profiler stack trace IDs as span attributes

JonasKunz:
Implements the final part of the universal profiling correlation: The IDs of CPU stacktraces are attached as OTel attributes to local root spans. In order to ensure that the spans are not sent before all their data has arrived, they are buffered and delayed for a configurable amount of time.

In a future iteration we hopefully won't need to configure this aspect: The profiler will tell the agent via the communication socket what delay is expected, so that the agent can configure it automatically.

Update:
This PR has been successfully tested against a real profiling host agent with the APM integration implemented.


universal-profiling-integration/src/main/java/co/elastic/otel/ProfilerSharedMemoryWriter.java
```diff
@@ -0,0 +1,121 @@
...
+import io.opentelemetry.api.trace.SpanContext;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.semconv.ResourceAttributes;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.charset.StandardCharsets;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class ProfilerSharedMemoryWriter {
```
JonasKunz, line 33: [For Review] this code isn't actually new, it has just been moved out of the SpanProcessor to keep things concise

jackshirazi (approved):

PR #163, opened 2024-03-07
buildkite: use VM with the installed tools

v1v:
### What

* Support custom VM images that contain the JDK17, gradle and the dependencies.
* Remove runtime installations.
* Use `.java-version` to store the current version based on major, as explained in https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#java-version-file

### Why

* Stability and reliability.
* Faster builds.

### Follow-up

- VM generation should be automatically recreated if the `.java-version` changes. Likely on a timer basis (daily)

### Test

Triggering manually against this particular feature branch (for such, I use the snapshot one):
- Snapshot [build](https://buildkite.com/elastic/elastic-otel-java-snapshot/builds/122)
- Release [build](https://buildkite.com/elastic/elastic-otel-java-release/builds/8).


.ci/snapshot.sh
```diff
@@ -23,12 +23,12 @@ echo $PATH
 java -version
 
 publishArg=''
-if [[ "$dry_run" == "true" ]] ; then
-    echo "--- Build and publish the snapshot :package: (dry-run)"
-    publishArg='publishAllPublicationsToDryRunRepository'
-else
+if [[ "$dry_run" == "false" ]] ; then
```
v1v, line 26: I prefer to play safe and do if `dry_run=false` is explicitly set. Otherwise, it's a dry-run mode.

.buildkite/hooks/pre-command
```diff
@@ -15,13 +15,19 @@ if [[ "$BUILDKITE_COMMAND" =~ .*"upload".* ]]; then
...
 
 if [ "$BUILDKITE_PIPELINE_SLUG" == "elastic-otel-java-release" ]; then
-    source .buildkite/hooks/prepare-release.sh
+  source .buildkite/hooks/prepare-release.sh
 fi
 
-# Run always
-source .buildkite/hooks/prepare-common.sh
+# Configure the java version
+JAVA_VERSION=$(cat .java-version)
```
amannocci, line 26:
```suggestion
JAVA_VERSION=$(cat .java-version | xargs | tr -dc '[:print:]')
```
We can be defensive about the `.java-version` file content.

.buildkite/hooks/pre-command
```diff
@@ -15,13 +15,19 @@ if [[ "$BUILDKITE_COMMAND" =~ .*"upload".* ]]; then
...
+JAVA_HOME="$HOME/.java/openjdk$JAVA_VERSION"
```
amannocci, line 27:
```suggestion
JAVA_HOME="$HOME/.java/openjdk${JAVA_VERSION}"
```

.buildkite/hooks/pre-command
```diff
@@ -15,13 +15,19 @@ if [[ "$BUILDKITE_COMMAND" =~ .*"upload".* ]]; then
...
+export JAVA_HOME
+PATH=$JAVA_HOME/bin:$PATH
```
amannocci, line 29:
```suggestion
PATH="${JAVA_HOME}/bin:${PATH}"
```

.java-version
```diff
@@ -0,0 +1 @@
+17
```
SylvainJuge, line 1: What is the exact syntax that we have to use here ? Also, which vendor version of the JDK is used here ? Does it refers to anything in the Elastic JVM catalog ?

v1v: > What is the exact syntax that we have to use here ?

In this case, only major versions.

> which vendor version of the JDK is used here ?

For GitHub actions, it uses
- `temurin`, see https://github.com/elastic/elastic-otel-java/blob/091edb229a7b2c6baf90331c9a605265dbeb98d8/.github/workflows/gradle-goal/action.yml#L14

For Buildkite, it uses:
- https://jvm-catalog.elastic.co/jdk and more precisely [openjdk](https://github.com/elastic/elastic-otel-java/blob/091edb229a7b2c6baf90331c9a605265dbeb98d8/.buildkite/hooks/prepare-common.sh#L26C10-L28)

SylvainJuge: Nevermind, I found the answer is included in this PR, so it's effectively using the "latest openjdk from Elastic JVM catalog".

```
  JAVA_URL=https://jvm-catalog.elastic.co/jdk
  JAVA_PKG="${JAVA_URL}/latest_openjdk_${JAVA_VERSION}_linux.tar.gz"
```

.ci/release.sh
```diff
@@ -23,12 +23,12 @@ echo $PATH
 java -version
 
 publishArg=''
-if [[ "$dry_run" == "true" ]] ; then
-    echo "--- Build and publish the release :package: (dry-run)"
-    publishArg='publishAllPublicationsToDryRunRepository'
-else
+if [[ "$dry_run" == "false" ]] ; then
```
SylvainJuge, line 26: Here the condition must be reversed (or the if/else blocks swapped).

.ci/snapshot.sh
SylvainJuge: Here do you mean it should default to dry-run and only use the dry-run when explicitly executed with `dry_run=false` ? Maybe updating the default value in the buildkite input pipeline would be relevant for consistency.

.github/workflows/build.yml
```diff
@@ -26,18 +26,15 @@ jobs:
...
-          distribution: temurin
-          java-version: 17
-          cache: gradle
-
       - name: Gradle build, test and check
-        uses: gradle/gradle-build-action@v3
+        uses: ./.github/workflows/gradle-goal
         with:
-          arguments: check
+          command: "./gradlew check"
```
SylvainJuge, line 32: [minor] maybe it would be worth to keep the `./gradlew` part handled in the `gradle-goal` workflow and only provide the gradle arguments here.

.ci/snapshot.sh
v1v: I meant that the release/snapshot should only run if `dry_run=false`.

> Maybe updating the default value in the buildkite input pipeline would be relevant for consistency.

That's already in the GitHub action
- https://github.com/elastic/elastic-otel-java/blob/091edb229a7b2c6baf90331c9a605265dbeb98d8/.github/workflows/release.yml#L86


Buildkite pipelines are not aimed to be ran manually, it's possible to use them, but I only did it to validate these changes. 


There is no concept of buildkit input values but build environment variables. Do you still want to default to `false`? It's possible to do it, but I dson't see the value now with my current proposal

.github/workflows/build.yml
v1v: I can do that

.ci/release.sh
v1v: I don't think so.

`dry-run==false` then do a release. Otherwise, dry-run is enabled.

This the way we can avoid surprises, such as:


```bash
dry_run='ups'


if [[ "$dry_run" == "true" ]] ; then
  publishArg='publishAllPublicationsToDryRunRepository'
else
  publishArg='publishToSonatype closeAndReleaseStagingRepository'
fi
...


It will run a release even when `dry_run` was not set to false but `ups`


I understand `dry-run` is a bit overkilling with a `false` validation but it's safer IMO

.github/workflows/build.yml
v1v: Wait, if I do so, then the gradle-goal will convert in `*nix` based. So windows won't be supported.

Let's keep it for now

.ci/snapshot.sh
SylvainJuge: Thanks for the clarification here, making the dry-run the default behavior here only applies to the buildkite pipeline when the env variable `dry_run` is not provided or has a value that is not `true` or `false`, and there is no need to also do the same at the GH actions workflow.

When the BK job is triggered from GH actions, the parameter is always provided with either `true` or `false` value.

.github/workflows/build.yml
SylvainJuge: Good for me, that's definitely something we can improve later.

amannocci (approved):

amannocci (approved):

PR #169, opened 2024-03-13
Fix and centralize publishing logic

### LINKED ISSUE #151: Build refactor: Move signing and publishing to convention in buildSrc
We currently copy-paste the buildscript configuring the signing and publication (e.g. to `inferred-spans`).
We should use a gradle convention plugin instead.

In general, we should try to remove as much logic as possible from the `subprojects {}` call in the root project, as this makes the build difficult to understand and [is not recommended](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration). We shoult stick to only doing very basic things which are guaranteed to be always the same for all projects, such as assigning the `groupid` and `version`.

JonasKunz:
Closes #151.

Also currently our release was kind of broken: we published `inferred-spans` as a library, but didn't publish the `common` project it depends on.

### DIFF
--- a/buildSrc/src/main/kotlin/elastic-otel.library-packaging-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/elastic-otel.library-packaging-conventions.gradle.kts
@@ -11,7 +11,6 @@ tasks {
 
       include("LICENSE")
       include("NOTICE")
-      include("licenses/**")
     }
   }
 
--- a/buildSrc/src/main/kotlin/elastic-otel.sign-and-publish-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/elastic-otel.sign-and-publish-conventions.gradle.kts
@@ -19,6 +19,10 @@ publishingConventions.artifactTasks.convention(listOf())
 
 afterEvaluate {
 
+  if (project.description == null || project.description!!.isBlank()) {
+    throw GradleException("Project description must be set to publish the project to maven central!")
+  }
+
   publishing {
 
     repositories {
@@ -53,6 +57,7 @@ afterEvaluate {
         }
 
         pom {
+
           name.set(project.description)
           description.set(project.description)
           url.set("https://github.com/elastic/elastic-otel-java")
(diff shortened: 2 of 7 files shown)


buildSrc/src/main/kotlin/elastic-otel.library-packaging-conventions.gradle.kts
```diff
@@ -11,7 +11,6 @@ tasks {
 
       include("LICENSE")
       include("NOTICE")
-      include("licenses/**")
```
JonasKunz, line 14: Our library project are *not* fat-jars: Their dependencies are not shipped with them, but listed in the pom.
Therefore I think it is wrong to include the licenses of those dependencies within the library jars.

SylvainJuge: I agree, that's definitely fair to not include them in this case.

buildSrc/src/main/kotlin/elastic-otel.sign-and-publish-conventions.gradle.kts
```diff
@@ -0,0 +1,91 @@
...
+        } else {
+          //publish just the artifacts (fat-jars).
+          for (task in artifactTasks) {
+            artifact(task)
+          }
+        }
+
+        pom {
+          name.set(project.description)
+          description.set(project.description)
```
SylvainJuge, line 57: [minor] Maybe that would be worth a later improvement but do we have default values for the project description ? Or maybe it's probably better to make this mandatory and throw an explicit error when value is not provided.

SylvainJuge (approved):

buildSrc/src/main/kotlin/elastic-otel.sign-and-publish-conventions.gradle.kts
JonasKunz: Good point, I'll add a check to throw an error if the description is unset.

JonasKunz: If the description is missing, gradle will now show an error at configuration time (so even without trying to publish).

PR #175, opened 2024-03-13
script upstream update

SylvainJuge:
Adds a simple (and brittle) shell script that helps updating upstream dependencies when there is a new release.

takes git ref as parameter and does a simple parsing of the upstream agent dependencies files, so it's not perfect nor robust but does the job for now until we need better.

For reference, updates in the upstream project seem to also be relying on sed commands, for example here:
- https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/scripts/update-sdk-version.sh
- https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/scripts/update-version.sh

### DIFF
--- a/.ci/update-upstream.sh
+++ b/.ci/update-upstream.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+set -x
+
+upstreamRef="${1:-main}"
+
+upstream_base_url="https://raw.githubusercontent.com/open-telemetry/opentelemetry-java-instrumentation/${upstreamRef}"
+project_root="$(dirname "${0}")/.."
+
+upstream_version() {
+    set +e
+    curl -s "${upstream_base_url}/${1}" \
+        | grep "val.*${2}.*=" \
+        | sed 's/.*=//' \
+        | tr -d '"' \
+        | tr -d '[:blank:]' \
+        || echo 'unknown'
+    set -e
+}
+
+upstreamAgentVersion="$(upstream_version 'version.gradle.kts' stableVersion)"
+upstreamAgentAlphaVersion="$(upstream_version 'version.gradle.kts' alphaVersion)"
+upstreamSdkVersion="$(upstream_version 'dependencyManagement/build.gradle.kts' otelSdkVersion)"
+upstreamContribVersion="$(upstream_version 'dependencyManagement/build.gradle.kts' otelContribVersion)"
+
+sed -i "s/^openTelemetrySdk = \".*/openTelemetrySdk = \"${upstreamSdkVersion}\"/" "${project_root}/gradle/libs.versions.toml"
+sed -i "s/^opentelemetryJavaagent = \".*/opentelemetryJavaagent = \"${upstreamAgentVersion}\"/" "${project_root}/gradle/libs.versions.toml"
+sed -i "s/^opentelemetryJavaagentAlpha = \".*/opentelemetryJavaagentAlpha = \"${upstreamAgentAlphaVersion}\"/" "${project_root}/gradle/libs.versions.toml"
+sed -i "s/^opentelemetryContribAlpha = \".*/opentelemetryContribAlpha = \"${upstreamContribVersion}\"/" "${project_root}/gradle/libs.versions.toml"
+
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -6,20 +6,23 @@ junit = "5.10.2"
 autoservice = "1.1.1"
 
 # otel SDK and API
+# updated from upstream agent with .ci/update-upstream.sh
 opentelemetrySdk = "1.35.0"
 
 # otel protocol (OTLP)
 opentelemetryProto = "1.1.0-alpha"
 
 # otel agent, those two should be updated together
+# updated from upstream agent with .ci/update-upstream.sh
 opentelemetryJavaagent = "2.1.0"
-opentelemetryJavaagentAlpha = "2.2.0-alpha-SNAPSHOT"
+opentelemetryJavaagentAlpha = "2.1.0-alpha"
 
 # otel semconv java generated bindings, needs to be kept in sync with version in otel agent
 opentelemetrySemconv = "1.23.1-alpha"
 
 # otel contrib
-opentelemetryContribAlpha = "1.34.0-alpha-SNAPSHOT"
+# updated from upstream agent with .ci/update-upstream.sh
+opentelemetryContribAlpha = "1.33.0-alpha"
 
 [libraries]
 


gradle/libs.versions.toml
```diff
@@ -6,20 +6,23 @@ junit = "5.10.2"
...
-opentelemetryJavaagentAlpha = "2.2.0-alpha-SNAPSHOT"
+opentelemetryJavaagentAlpha = "2.1.0-alpha"
 
 # otel semconv java generated bindings, needs to be kept in sync with version in otel agent
 opentelemetrySemconv = "1.23.1-alpha"
 
 # otel contrib
-opentelemetryContribAlpha = "1.34.0-alpha-SNAPSHOT"
+# updated from upstream agent with .ci/update-upstream.sh
+opentelemetryContribAlpha = "1.33.0-alpha"
```
SylvainJuge, line 25: [for reviewer] here the dependabot update messed up, and we might later configure to ignore those dependency and only update manually through our script.

gradle/libs.versions.toml
```diff
@@ -6,20 +6,23 @@ junit = "5.10.2"
...
 opentelemetrySdk = "1.35.0"
 
 # otel protocol (OTLP)
 opentelemetryProto = "1.1.0-alpha"
 
 # otel agent, those two should be updated together
+# updated from upstream agent with .ci/update-upstream.sh
 opentelemetryJavaagent = "2.1.0"
-opentelemetryJavaagentAlpha = "2.2.0-alpha-SNAPSHOT"
+opentelemetryJavaagentAlpha = "2.1.0-alpha"
```
SylvainJuge, line 18: [for reviewer] this is exactly the kind of inconsistency that we get with current updates with dependabot (which also messes up with snapshots for now).

.github/dependabot.yml
```diff
@@ -22,3 +22,9 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 10
+    ignore:
+      # upstream dependencies that need to be updated in sync and with manual control
```
SylvainJuge, line 26: [for reviewer] those dependencies will need to be manually updated by using the script or through manual edits (for example the current 2.1.0 release does not contain any dependency on the contrib that we could reuse, so we have to do it manually for now).

gradle/libs.versions.toml
```diff
@@ -6,20 +6,23 @@ junit = "5.10.2"
 autoservice = "1.1.1"
 
 # otel SDK and API
+# updated from upstream agent with .ci/update-upstream.sh
```
JonasKunz, line 9: I jsut noticed that the agent bom includes the otel bom:

https://repo1.maven.org/maven2/io/opentelemetry/instrumentation/opentelemetry-instrumentation-bom/2.1.0/opentelemetry-instrumentation-bom-2.1.0.pom

We should be able to get rid of the SDK version in our projects and instead have it managed by the opentelemetry-instrumentation-bom. Could you try this?

SylvainJuge: I think that could be a good option here, I'll try to reuse this BOM dependency so it's one less moving part for the SDK/API.

gradle/libs.versions.toml
```diff
@@ -6,20 +6,23 @@ junit = "5.10.2"
...
 opentelemetrySdk = "1.35.0"
 
 # otel protocol (OTLP)
 opentelemetryProto = "1.1.0-alpha"
 
 # otel agent, those two should be updated together
+# updated from upstream agent with .ci/update-upstream.sh
 opentelemetryJavaagent = "2.1.0"
```
JonasKunz, line 17: The opentelemetry-instrumentation bom is transitively imported form the alpha-bom:
Check the bottom of https://repo1.maven.org/maven2/io/opentelemetry/instrumentation/opentelemetry-instrumentation-bom-alpha/2.1.0-alpha/opentelemetry-instrumentation-bom-alpha-2.1.0-alpha.pom

This means we can get rid of explicitly listing and syncing that one too

gradle/libs.versions.toml
```diff
@@ -5,36 +5,29 @@ spotless = "6.25.0"
...
+opentelemetryJavaagentAlpha = "2.1.0-alpha"
 
 # otel contrib
-opentelemetryContribAlpha = "1.34.0-alpha-SNAPSHOT"
+# updated from upstream agent with .ci/update-upstream.sh
+opentelemetryContribAlpha = "1.33.0-alpha"
 
 [libraries]
 
+# For now those aren't resolved from bom for 'upstreamAgent', thus we have to keep them explicit
```
SylvainJuge, line 22: [for reviewer] we have to keep those two as the 'upstreamAgent' does not seem to work with `platform` dependencies, thus we need to keep those for now, each of them is only used in a single place though, this is the only reason we have to keep a version on the non-alpha otherwise both could be resolved from the `-alpha` bom.

JonasKunz: It does actually work, you just need to enforce the `platform` for the `upstreamAgent` configuraiton aswell:
At the moment we only did it for the `implementation`-configuration.

.ci/update-upstream.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
JonasKunz, line 1: This script isn't used by the CI (right?) but is placed in the `.ci` folder. I'd rather move it to the `gradle` folder so that it is next to the version cataloge.

JonasKunz (approved):

.ci/update-upstream.sh
SylvainJuge: Good point, I think we'll run it either manually to switch to a snapshot of whenever there is a new upstream release, so it's better to move it where you've suggested.

.ci/update-upstream.sh
```diff
@@ -0,0 +1,26 @@
...
+        | tr -d '[:blank:]' \
+        || echo 'unknown'
+    set -e
+}
+
+upstreamAgentAlphaVersion="$(upstream_version 'version.gradle.kts' alphaVersion)"
+upstreamContribVersion="$(upstream_version 'dependencyManagement/build.gradle.kts' otelContribVersion)"
+
+sed -i "s/^opentelemetryJavaagentAlpha = \".*/opentelemetryJavaagentAlpha = \"${upstreamAgentAlphaVersion}\"/" "${project_root}/gradle/libs.versions.toml"
+sed -i "s/^opentelemetryContribAlpha = \".*/opentelemetryContribAlpha = \"${upstreamContribVersion}\"/" "${project_root}/gradle/libs.versions.toml"
```
JonasKunz, line 25:
```suggestion
sed -i '' -e "s/^opentelemetryJavaagentAlpha = \".*/opentelemetryJavaagentAlpha = \"${upstreamAgentAlphaVersion}\"/" "${project_root}/gradle/libs.versions.toml"
sed -i '' -e "s/^opentelemetryContribAlpha = \".*/opentelemetryContribAlpha = \"${upstreamContribVersion}\"/" "${project_root}/gradle/libs.versions.toml"
```

The original sed command fails on mac:
`sed: 1: "./.ci/../gradle/libs.ve ...": invalid command code .`

PR #177, opened 2024-03-14
Add field-backed storage for SpanValues

JonasKunz:
This PR adds a new `$elasticSpanValues` field to the `SdkSpan` class of our agent distro at build time.
This field will then be used for more efficient storage of `SpanValues` instead of using a `WeakConcurrentHashMap`.


testing/integration-tests/agent-internals/build.gradle.kts
```diff
@@ -0,0 +1,9 @@
+plugins {
+    id("elastic-otel.test-with-agent-conventions")
+}
+
+tasks.test {
+  jvmArgs(
+    //"-Dotel.javaagent.debug=true"
```
SylvainJuge, line 7: [minor] looks like leftover from debugging session, maybe we could have a way to consistently enable debug logging when running tests, or simply rely on the value of `OTEL_JAVAAGENT_DEBUG=true` which is also automatically enabled when debugging smoke tests.

testing/integration-tests/agent-internals/src/test/java/SpanValueTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    Tracer tracer = GlobalOpenTelemetry.get().getTracer("my-tracer");
+    Span bridgeSpan = tracer.spanBuilder("s1").startSpan();
+
+    // Move from the bridged span to the agent internal Span instance (of the Agent's SDK).
+    Object agentSpan = readFieldValue(bridgeSpan, "agentSpan");
+    assertThat(agentSpan.getClass().getSimpleName()).isEqualTo("SdkSpan");
+
+    // From that Classloader we should be able to find and create SpanValues
+    Class<?> SpanValueClass =
```
SylvainJuge, line 44:
```suggestion
    Class<?> spanValueClass =
```

testing/integration-tests/agent-internals/src/test/java/SpanValueTest.java
```diff
@@ -0,0 +1,86 @@
...
+        Class.forName(
+            "co.elastic.otel.common.SpanValue", true, agentSpan.getClass().getClassLoader());
+    Class<?> ReadableSpanInterface =
```
SylvainJuge, line 47:
```suggestion
    Class<?> readableSpanInterface =
```

SylvainJuge (approved):

common/src/main/java/io/opentelemetry/sdk/trace/FieldBackedSpanValueStorageProvider.java
```diff
@@ -0,0 +1,104 @@
...
+      setter = AtomicReferenceFieldUpdater.newUpdater(SdkSpan.class, Object.class, FIELD_NAME);
+
+      logger.log(Level.FINE, "Using field-backed storage for SpanValues", FIELD_NAME);
+      resultInstance = new FieldBackedSpanValueStorageProvider();
+
+    } catch (NoSuchFieldException e) {
+      logger.log(
+          Level.FINE,
+          "Using map-backed storage for SpanValues because Field '{0}' does not exist on SdkSpan",
+          FIELD_NAME);
```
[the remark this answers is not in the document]
JonasKunz: This is intentionally on debug level because this is the expected case when running our extensions outside of the agent with a vanilla SDK. If anything else goes wrong, we throw an error in the catch clause below.

At the same time we should catch problems with our code modification with the integration test I added in this PR.

testing/integration-tests/agent-internals/build.gradle.kts
JonasKunz: > enable debug logging when running tests

We could do this via the `test-with-agent-conventions` if you want to. I just felt like having debug always on on everything is a bit noisy.

I intentionally left it there as a comment to allow easy enabling of the debug logs.

> simply rely on the value of OTEL_JAVAAGENT_DEBUG=true which is also automatically enabled when debugging smoke tests.

What do you mean with `rely on the value of OTEL_JAVAAGENT_DEBUG=true` ? This is not a smoke-test, therefore the logic implemented there doesn't have any effect in these tests.

common/src/main/java/io/opentelemetry/sdk/trace/FieldBackedSpanValueStorageProvider.java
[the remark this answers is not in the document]
SylvainJuge: That's a very good point, I forgot about usage of this with just an extension, so the SDK code modification would not be applied and then the fallback solution is needed to have something working (albeit with few compromises on performance).

PR #179, opened 2024-03-14
Split building and testing in workflow

JonasKunz:
This has two benefits:
 * Allows us to easily introduce a matric-job for testing various JVM versions without always having to recompile
 * If the workflow fails, it will be immediately clear if it was due to compilation/checkStyle or tests
 
 Preparation for #180.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,7 @@ jobs:
         with:
           name: working-dir-build-cache
           path: ./wdarch/working-dir-build-cache.tar.gz
+          retention-days: 1 # No need to waste space when this artifact is only used as part of the build process
 
       - name: Agent artifact
         uses: actions/upload-artifact@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,11 +32,14 @@ jobs:
           # We also compile the test-classes, even though we are skipping the tests
           command: "./gradlew build testClasses -x test"
 
+      - name: Create working dir archive
+        run: "tar --exclude='./working-dir-build-cache.tar.gz' -czf ./working-dir-build-cache.tar.gz ."
+
       - name: Cache working directory with build results
         uses: actions/upload-artifact@v3
         with:
           name: working-dir-build-cache
-          path: ./
+          path: ./working-dir-build-cache.tar.gz
 
       - name: Agent artifact
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -19,32 +19,60 @@ jobs:
...
-        uses: ./.github/workflows/gradle-goal
-        with:
-          command: "./gradlew check"
-      - name: Store test results
-        if: success() || failure()
+      # We first tar the working directory and then upload it
+      # We do this because uploading directories with many files directly as artifacts is very slow
+      - name: Create working dir archive
+        run: "mkdir ./wdarch && tar --exclude='./wdarch' -czf ./wdarch/working-dir-build-cache.tar.gz ."
+      - name: Cache working directory with build results
```
v1v, line 39: Maybe `v4` can be faster? See https://github.com/actions/upload-artifact?tab=readme-ov-file#usage

In this case you can case use v4 since it's not a matrix job uploading artifacts

v1v (approved):
LGTM with just a minor nit

.github/workflows/build.yml
JonasKunz: Just did a test, v4 seems to be significantly faster. It is even faster than my tar solution:
image

I'll update the PR ans switch to the `@v4` variants

JonasKunz: Unfortunatly the `upload-artifact` action doesn't preserve permissions:
```
/home/runner/work/_temp/951c9356-ece7-4462-9b6c-2985a0d9fb9b.sh: line 1: ./gradlew: Permission denied
```

As suggested by the [readme](https://github.com/actions/upload-artifact?tab=readme-ov-file#permission-loss), I'll stay with taring but change the comment explaining the reasoning

v1v: Oh dear! I missed that. Tath's a pain in the heck

v1v: I wonder if you only need to store the `build` folder? If so, will the permissions won't be a problem?

JonasKunz: You would need to store the outputs of every subproject which usually should go into the build folder. In addition I think gradle stores the hashes of the build contents somewhere, but I'm not sure about where.

Also, you'd need to checkout again and merge things back. I think just uploading the entire working dir and downloading it again is a lot simpler and less error-prone than checking out the repo again and trying to merge things back together.

SylvainJuge (approved):

PR #183, opened 2024-03-14
fix update script regex

SylvainJuge:
Fix regex for mac.

Needs validation from @JonasKunz on a mac.


JonasKunz (approved):
Tested it manually, works as expected.

PR #187, opened 2024-03-26
Allow testing for different JVM versions and implementations

### LINKED ISSUE #180: Test on various JVM versions on CI
Currently we only run tests on Temurin JDK 17.
This means we accidentally could have Java 8 incompatibilities in our code.
In addition, we have some tests related to virtual threads which require Java 21 to run.

We should change the testing on CI to run on different JVM versions and distros.

JonasKunz:
Closes #180 .

I've intentionally used the same approach and property names like the upstream agent, so that we don't get confused when working on both:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/9d94b4fd7e5ebcf227e29a6d4e1c343621c46438/conventions/src/main/kotlin/otel.java-conventions.gradle.kts#L375

### DIFF
--- a/.github/workflows/test-reporter.yml
+++ b/.github/workflows/test-reporter.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - uses: elastic/apm-pipeline-library/.github/actions/test-report@current
         with:
-          artifact: test-results            # artifact name
-          name: JUnit Tests                 # Name of the check run which will be created
+          artifact: /test-results-(.*)/     # artifact name
+          name: 'Test Results $1'           # Name of the check run which will be created
           path: "**/*.xml"                  # Path to test results (inside artifact .zip)
           reporter: java-junit              # Format of test results
--- a/smoke-tests/build.gradle
+++ b/smoke-tests/build.gradle
@@ -2,7 +2,7 @@ import jdk.internal.joptsimple.internal.Strings
 
 // TODO : convert to kotlin for consistency
 plugins {
-  id "java"
+  id("elastic-otel.java-conventions")
   alias catalog.plugins.taskinfo
 }
 
@@ -12,6 +12,10 @@ configurations {
     canBeResolved = true
     canBeConsumed = false
   }
+}
+
+compileTestJava {
+  options.release.set(17)
 }
 
 dependencies {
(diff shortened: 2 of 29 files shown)


smoke-tests/build.gradle
```diff
@@ -14,6 +14,10 @@ configurations {
   }
 }
 
+compileTestJava {
+  options.release.set(17)
```
JonasKunz, line 18: [For Reviwer] Note that it is not strictly required that all tests are Java 8 compatible.
It is possible to simply override the minimum testing java version on a per-project basis like here.
In this case the smoke-tests won't be executed for Java versions older than 17

.github/workflows/test-reporter.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     steps:
       - uses: elastic/apm-pipeline-library/.github/actions/test-report@current
         with:
-          artifact: test-results            # artifact name
-          name: JUnit Tests                 # Name of the check run which will be created
+          artifact: /test-results-(.*)/     # artifact name
+          name: 'Test Results $1'           # Name of the check run which will be created
```
JonasKunz, line 23: [For Reviewer]: The same approach was already used and tested in the `elastic-apm-agent`
https://github.com/elastic/apm-agent-java/blob/main/.github/workflows/test-reporter.yml

smoke-tests/build.gradle
SylvainJuge: [question] how does those smoke tests get skipped and reported when executing tests on java8 ?
- are they absent, failed (but ignored) or marked as skipped in the test report ?
- do we have to do anything specific to make them ignored (or executed) depending on the JVM used to run them ? Is there any kind of opt in/out or is it transparently managed by the toolchain in gradle ?

JonasKunz: The corresponding gradle `test` task will be simply disabled, so the tests won't be loaded by the target JVM and therefore don't appear in any state in the test reports.

See this line for the logic:
https://github.com/elastic/elastic-otel-java/pull/187/files#diff-8c76762546ca458066ae00e3880c724d446b6b0cad63cf585c71a9d27b91a0fbR19

If you write your tests Java 8 compilable, but only want them to run on specific JVMs you can use the JUnit annotations on your tests instead. This case is just for where we want to use libraries or Language features in tests which cannot be compiled to Java 8. In this case, the smoke tests use Text-blocks for readability and it doesn't make sense to run the smoketests on older versions, because the actual testing happens inside containers anyway.

If we wanted to be super clean, we could introduce a separate task + workflow job for the smoke tests, but this is not really worth it now IMO.

SylvainJuge (approved):

PR #188, opened 2024-03-27
Fix test reporting

JonasKunz:
Looks like #187 broke the test reporting workflow:
```
Error: Response code 400 (Authentication information is not given in the correct format. Check the value of Authorization header.)
```
https://github.com/elastic/elastic-otel-java/actions/runs/8453636169/job/23156761898#step:2:42

I assume it is due to the change of the `upload-artifact` action from `@v3` to `@v4`.


SylvainJuge:
I remember there was quite a few reasons for the robots team to not use the v4 version, does this one is already a known reason for not using it ? Maybe there is a list of related issues somewhere already.

SylvainJuge (approved):

JonasKunz:
I think this was one of the known reasons.

PR #190, opened 2024-03-28
Fix flaky InferredSpansTest

JonasKunz:


JonasKunz:
Ran the tests five times without a failure, so looks like the fix does it's job.

SylvainJuge (approved):

PR #198, opened 2024-04-09
Fix tests due to new config option in old agent

JonasKunz:


JonasKunz:
Closes in favour of #199

PR #201, opened 2024-04-11
Fix inferred span cycles

JonasKunz:


JonasKunz:
@AlexanderWert confirmed with a manual test that this fixes the problem

AlexanderWert (approved):

PR #204, opened 2024-04-15
Handle profiler registration messages

JonasKunz:
Implements https://github.com/elastic/apm/pull/853:
 * If the profiler provides a `host.id` it will be added to the spans
 * The profiler can now tell the extension how long to wait for the stacktraces for spans to arrive
 * Added option for the extension to remain inactive until a profiler is actually detected

### DIFF
--- a/universal-profiling-integration/src/main/java/co/elastic/otel/SpanProfilingSamplesCorrelator.java
+++ b/universal-profiling-integration/src/main/java/co/elastic/otel/SpanProfilingSamplesCorrelator.java
@@ -53,7 +53,7 @@ public class SpanProfilingSamplesCorrelator {
   private final RingBuffer<DelayedSpan> delayedSpans;
   private final PeekingPoller<DelayedSpan> delayedSpansPoller;
 
-  private volatile long spanDelayNanos;
+  private volatile long spanBufferDurationNanos;
   private volatile boolean shuttingDown = false;
 
   private final WriterReaderPhaser shutdownPhaser = new WriterReaderPhaser();
@@ -64,7 +64,7 @@ public class SpanProfilingSamplesCorrelator {
       long initialSpanDelayNanos,
       Consumer<ReadableSpan> sendSpan) {
     this.nanoClock = nanoClock;
-    this.spanDelayNanos = initialSpanDelayNanos;
+    this.spanBufferDurationNanos = initialSpanDelayNanos;
     this.sendSpan = sendSpan;
 
     bufferCapacity = nextPowerOf2(bufferCapacity);
@@ -79,8 +79,11 @@ public class SpanProfilingSamplesCorrelator {
     delayedSpansPoller = new PeekingPoller<>(nonPeekingPoller, DelayedSpan::new);
   }
 
-  public void setSpanDelayNanos(long delay) {
-    spanDelayNanos = delay;
+  public void setSpanBufferDurationNanos(long nanos) {
+    if (nanos < 0) {
+      throw new IllegalArgumentException("nanos must be positive but was " + nanos);
+    }
+    spanBufferDurationNanos = nanos;
   }
 
   public void onSpanStart(ReadableSpan span, Context parentCtx) {
@@ -100,7 +103,7 @@ public class SpanProfilingSamplesCorrelator {
 
     long criticalPhaseVal = shutdownPhaser.writerCriticalSectionEnter();
     try {
-      if (spanDelayNanos == 0 || shuttingDown) {
+      if (spanBufferDurationNanos == 0 || shuttingDown) {
         correlateAndSendSpan(span);
         return;
       }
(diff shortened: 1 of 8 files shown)


universal-profiling-integration/src/main/java/co/elastic/otel/SpanProfilingSamplesCorrelator.java
```diff
@@ -78,6 +79,10 @@ public SpanProfilingSamplesCorrelator(
     delayedSpansPoller = new PeekingPoller<>(nonPeekingPoller, DelayedSpan::new);
   }
 
+  public void setSpanDelayNanos(long delay) {
+    spanDelayNanos = delay;
```
SylvainJuge, line 83: [minor] maybe guard against negative value

universal-profiling-integration/src/main/java/co/elastic/otel/UniversalProfilingProcessor.java
```diff
@@ -80,6 +84,8 @@ public class UniversalProfilingProcessor extends AbstractChainingSpanProcessor {
   // Visibile for testing
```
SylvainJuge, line 84:
```suggestion
  // Visible for testing
```

universal-profiling-integration/src/main/java/co/elastic/otel/UniversalProfilingProcessor.java
```diff
@@ -99,9 +105,18 @@ public static UniversalProfilingProcessorBuilder builder(SpanProcessor next, Res
                 + " There must be at most one processor of this type active at a time!");
       }
 
+      long initialSpanDelay;
+      if (activeOnlyAfterProfilerRegistration) {
+        initialSpanDelay = 0; // do not buffer spans until we know that a profiler is running
+        tlsPropagationActive = false;
+      } else {
+        initialSpanDelay = INITIAL_SPAN_DELAY_NANOS; // delay conservatively to not miss any data
```
SylvainJuge, line 113: [question] from this I understand that "span delay" here means the time duration for which the spans are buffered, is that correct ? If so, then probably renaming this to "span buffering duration" or similar seems more appropriate.

universal-profiling-integration/src/main/java/co/elastic/otel/UniversalProfilingProcessor.java
```diff
@@ -200,10 +215,12 @@ public boolean isEndRequired() {
   @Nullable
   private void onContextChange(@Nullable Context previous, @Nullable Context next) {
     try {
-      Span oldSpan = safeSpanFromContext(previous);
-      Span newSpan = safeSpanFromContext(next);
-      if (oldSpan != newSpan && !oldSpan.getSpanContext().equals(newSpan.getSpanContext())) {
-        ProfilerSharedMemoryWriter.updateThreadCorrelationStorage(newSpan);
+      if (tlsPropagationActive) {
```
SylvainJuge, line 218: [nit] maybe moving this check outside of the try/catch block with an early return to remove a few curly braces.

universal-profiling-integration/src/main/java/co/elastic/otel/UniversalProfilingProcessor.java
```diff
@@ -249,6 +268,20 @@ private void consumeProfilerMessages() {
...
+  private void handleMessage(ProfilerRegistrationMessage message) {
+    log.log(
+        Level.FINE,
+        "Received profiler registration message! host.id is {0} and the span delay is {1} ms",
+        new Object[] {message.getHostId(), message.getSamplesDelayMillis()});
+
+    tlsPropagationActive = true;
+    long spanDelayNanos =
+        Duration.ofMillis(message.getSamplesDelayMillis() + POLL_FREQUENCY_MS).toNanos();
+    correlator.setSpanDelayNanos(spanDelayNanos);
```
SylvainJuge, line 280: Here it's a good place where the "delay" provided by the profiler get mixed with the known polling frequency to produce a "time buffer" for which we need to buffer spans.

Also, would it be relevant to have the profiler also provide the sampling frequency rather than hard-coding it in the agent or is it something very unlikely to change on the profiler side ?

universal-profiling-integration/src/test/java/co/elastic/otel/UniversalProfilingProcessorTest.java
```diff
@@ -239,37 +247,36 @@ private String readUtf8Str(ByteBuffer buffer) {
...
-      if (tls != null) {
-        tls.order(ByteOrder.nativeOrder());
-        assertThat(tls.getChar(0)).isEqualTo((char) 1); // layout-minor-version
-        assertThat(tls.get(2)).isEqualTo((byte) 1); // valid byte
-      }
+  static void checkTlsIs(Span span, Span localRoot) {
+    ByteBuffer tls = JvmtiAccessImpl.createThreadProfilingCorrelationBufferAlias(TLS_STORAGE_SIZE);
+    if (tls != null) {
+      tls.order(ByteOrder.nativeOrder());
+      assertThat(tls.getChar(0)).isEqualTo((char) 1); // layout-minor-version
```
SylvainJuge, line 256: [nit] use `.describedAs("layout-minor-version")` to describe the assertion instead of comment.

SylvainJuge (approved):

universal-profiling-integration/src/main/java/co/elastic/otel/UniversalProfilingProcessor.java
JonasKunz: Correct, I'll rename it.

universal-profiling-integration/src/main/java/co/elastic/otel/UniversalProfilingProcessor.java
JonasKunz: The `POLL_FREQUENCY_MS` is actually unrelated to the sampling frequency of the profiler.
This duration just defined how frequently we look into the profiling socket to check for messages.

Therefore I don't see any need to make this configurable right now. If the need arises, we can add it later.

PR #205, opened 2024-04-15
Publish features as pre-packaged opentelemetry agent extension

### LINKED ISSUE #203: Publish features as vanilla agent extension
We currently make our distro features usable in two ways:
 * By using our opentelemetry agent distro
 * By setting up the SDK manually and adding our features as SDK extensions

The goal of this issue is to allow another way: Use the vanilla opentelemetry agent, but with our features as an extension jar for the existing agent extension mechanism. This requires the extensions to be bundled and published as a fat-jar.

Note that the agent extension jar should not set the distro name and version.

JonasKunz:
Closes #203 .


universal-profiling-integration/src/test/java/co/elastic/otel/UniversalProfilingProcessorTest.java
```diff
@@ -514,6 +514,7 @@ void shutdownFlushesBufferedSpans() {
 
       try (OpenTelemetrySdk sdk = initSdk(builder -> builder.clock(() -> 0L))) {
         sendProfilerRegistrationMsg(1, "hostid");
+        processor.pollMessagesAndFlushPendingSpans();
```
JonasKunz, line 517: This change fixes an actually unrelated flakyness: The test assumes that the profielr registration message has been consumed, but this isn't guaranteed by the background polling. For this reason we have to force the span processor to explicitly poll and process the messages.

testing/integration-tests/inferred-spans-test/build.gradle.kts
```diff
@@ -8,7 +8,7 @@ dependencies {
   testImplementation("io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations")
 }
 
-tasks.test {
+tasks.withType<Test>() {
   jvmArgs(
     //"-Dotel.javaagent.debug=true",
     "-Delastic.otel.inferred.spans.enabled=true",
```
jackshirazi, line 14: Did you test that this still works with the property on/off?

jackshirazi (approved):

testing/integration-tests/inferred-spans-test/build.gradle.kts
JonasKunz: What do you mean with property on off?
In case you are targeting at whether the tests actually test what they should: yes they do. This test actually uncovered in a previous iteration that the extension wasn't wokring correctly because the service files wouldn't get merged.

custom/src/main/java/co/elastic/otel/ElasticDistroResourceProvider.java
```diff
@@ -31,13 +31,21 @@ public class ElasticDistroResourceProvider implements ResourceProvider {
...
   public Resource createResource(ConfigProperties configProperties) {
-    return AgentVersion.VERSION == null
-        ? Resource.empty()
-        : Resource.create(
-            Attributes.of(
-                ElasticAttributes.TELEMETRY_DISTRO_NAME,
-                "elastic",
-                ElasticAttributes.TELEMETRY_DISTRO_VERSION,
-                AgentVersion.VERSION));
+    if (AgentVersion.VERSION == null) {
```
SylvainJuge, line 34: [question] should we provide a way to identify the extension version and name using maybe non-semconv attributes ? on the server/storage side it's often tricky to rely on "feature detection" based only on data shape or contents.

PR #215, opened 2024-04-17
Universal profiling integration autoconfig and inclusion in distro

JonasKunz:
Closes #45

PR #218, opened 2024-04-22
Fix agentextension publishing and LocalRootSpan bug in agentextension

JonasKunz:
The release currently failed due to a gradle cautionary check, which is a false-positive in this case:
```
* What went wrong:
Execution failed for task ':agentextension:publishMavenPublicationToSonatypeRepository'.
> Failed to publish publication 'maven' to repository 'sonatype'
   > Artifact elastic-otel-agentextension-0.1.0.jar wasn't produced by this build.
```

I've fixed the issue by reusing the same logic for publishing a shadow-jar we use for publishing our distro agent.

Also this PR includes two bug-fixes I forgot about which are also included in #215 . Without those, the agent-extension will not work properly.

### DIFF
--- a/common/src/main/java/co/elastic/otel/common/LocalRootSpan.java
+++ b/common/src/main/java/co/elastic/otel/common/LocalRootSpan.java
@@ -79,6 +79,19 @@ public class LocalRootSpan {
     }
   }
 
+  /** See {@link LocalRootSpan#getFor(Span)}. */
+  @Nullable
+  public static ReadableSpan getFor(ReadableSpan span) {
+    Object rootSpanVal = localRoot.get(span);
+    if (rootSpanVal == LOCAL_ROOT_MARKER) {
+      return span;
+    }
+    if (rootSpanVal == INFERRED_SPAN_UNKNOWN_ROOT_MARKER) {
+      return null;
+    }
+    return (ReadableSpan) rootSpanVal;
+  }
+
   /**
    * If the provided span is a local root span, itself is returned. Otherwise, returns the
    * (transitive) parent which is a local root.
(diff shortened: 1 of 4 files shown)


common/src/main/java/co/elastic/otel/common/LocalRootSpan.java
```diff
@@ -93,31 +106,14 @@ public static void onSpanStart(ReadableSpan startedSpan, Context parentContext)
...
-  /** See {@link LocalRootSpan#getFor(Span)}. */
-  @Nullable
-  public static Span getFor(ReadWriteSpan span) {
-    return resolveRoot(span, localRoot.get(span));
-  }
-
-  @Nullable
-  private static Span resolveRoot(Object span, Object rootSpanVal) {
-    if (rootSpanVal == LOCAL_ROOT_MARKER) {
-      return (Span) span;
```
JonasKunz, line 115: This cast would cause an exception if the provided span is a `MutableSpan`. That's why I changed `LocalRootSpan` to operate on `ReadableSpan` instead of `Span`.

common/src/main/java/co/elastic/otel/common/SpanValueStorageProvider.java
```diff
@@ -21,11 +21,27 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
 public interface SpanValueStorageProvider {
 
+  Logger logger = Logger.getLogger(SpanValueStorageProvider.class.getName());
+
   static SpanValueStorageProvider get() {
+    try {
```
JonasKunz, line 33: `FieldBackedSpanValueStorageProvider` directly references `SdkSpan`. If they are in different class loaders, this causes an `IllegalAccessError` here, which we didn't account for yet.

jackshirazi (approved):

PR #223, opened 2024-04-23
github-actions: enable provenance

v1v:
## What does this PR do?

Enable the provenance for the jar file generated in Buildkite using the [GitHub provenance action](https://github.com/github-early-access/generate-build-provenance).

User `artifacts.tar` to store all the artifacts that are produced (uses https://github.com/elastic/elastic-otel-java/pull/226)

## Test

I created a test feature branch and triggered the snapshot workflow:
- https://github.com/elastic/elastic-otel-java/actions/runs/8816874378

See https://github.com/elastic/elastic-otel-java/attestations 

## Issues

Similar to https://github.com/elastic/apm-agent-java/pull/3594

### DIFF
--- a/.ci/snapshot.sh
+++ b/.ci/snapshot.sh
@@ -40,3 +40,6 @@ if [[ "$dry_run" == "true" ]] ; then
     echo "--- Archive the dry-run repository :package: (dry-run)"
     tar czvf ./build/dry-run-maven-repo.tgz -C ./build/dry-run-maven-repo/ . | tee snapshot.txt
 fi
+
+echo "--- Archive the build folders with jar files"
+find . -type d -name build -exec find {} -name '*.jar' -print0 \; | xargs -0 tar -cvf "${TARBALL_FILE:-dist.tar}"
--- a/.ci/snapshot.sh
+++ b/.ci/snapshot.sh
@@ -42,4 +42,5 @@ if [[ "$dry_run" == "true" ]] ; then
 fi
 
 echo "--- Archive the build folders with jar files"
-find . -type d -name build -exec find {} -name '*.jar' -print0 \; | xargs -0 tar -cvf "${TARBALL_FILE:-dist.tar}"
+./gradlew printPublishedCodeArtifacts -q | tee artifacts.list
+tar -cvf "${TARBALL_FILE:-artifacts.tar}" -T artifacts.list
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,7 +74,7 @@ jobs:
       contents: write
       id-token: write
     env:
-      TARBALL_FILE: dist.tar
+      TARBALL_FILE: artifacts.tar
     steps:
       - id: buildkite
         name: Run Release
@@ -91,7 +91,7 @@ jobs:
           buildEnvVars: |
             ref=${{ inputs.ref }}
             dry_run=${{ inputs.dry_run || 'false' }}
-            TARBALL_FILE=dist.tar
+            TARBALL_FILE=${{ env.TARBALL_FILE }}
 
       - uses: actions/download-artifact@v3
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,6 +73,8 @@ jobs:
     permissions:
       contents: write
       id-token: write
+    env:
+      TARBALL_FILE: dist.tar
     steps:
       - id: buildkite
         name: Run Release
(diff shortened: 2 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -81,10 +84,21 @@ jobs:
           pipeline: elastic-otel-java-release
           waitFor: true
           printBuildLogs: false
+          artifactName: releases
```
SylvainJuge, line 87: [question] does it means that we only have one artifact (jar) file generated for the whole project ?, in practice we have more than one so do we have to do the same for each jar generated by the project ?

v1v: > [question] does it means that we only have one artifact (jar) file generated for the whole project ?,

Good question, I probably was wrong when I though that the only generated file was the one in https://github.com/elastic/elastic-otel-java/blob/6ed4c262282a1c16ba3991d2d4b13044bcc3259e/.buildkite/release.yml#L11 or files


>  in practice we have more than one so do we have to do the same for each jar generated by the project ?

the provenance accepts glob patterns, so it can be anything. At the moment, I only used `agent/build/libs/elastic-otel-javaagent-*.jar"` but if there are more jar files, please let me know and I can adjust the attestation to be done in all the files.

SylvainJuge: Currently we have 4 of them, and only the "agent" was covered in buildkite, they are already published in the snapshots as we can see here: https://oss.sonatype.org/service/local/repositories/snapshots/content/co/elastic/otel/.

I've opened https://github.com/elastic/elastic-otel-java/pull/224 to fix that. Once this is merged then you should be able to add the others.

v1v: Awesome, thanks @SylvainJuge

SylvainJuge: #224 has just been merged, you are good to go @v1v .

v1v: https://github.com/elastic/elastic-otel-java/pull/223/commits/2fa7c2aa17dc934d29bf19728c4d13c8dfd8cd39 is the proposal to search for all `jar` files in the `build` folders, then there is no need to maintain the list of artifacts.

reakaleek (approved):

.ci/snapshot.sh
```diff
@@ -40,3 +40,6 @@ if [[ "$dry_run" == "true" ]] ; then
     echo "--- Archive the dry-run repository :package: (dry-run)"
     tar czvf ./build/dry-run-maven-repo.tgz -C ./build/dry-run-maven-repo/ . | tee snapshot.txt
 fi
+
+echo "--- Archive the build folders with jar files"
+find . -type d -name build -exec find {} -name '*.jar' -print0 \; | xargs -0 tar -cvf "${TARBALL_FILE:-dist.tar}"
```
SylvainJuge, line 45: we often use github actions artifacts as a way to provide a direct download link for the snapshot versions when doing support, so using a `.zip` file would be better for non *nix users.

Also, the list of artifacts included is broader than what is necessary:
- not all `.jar` files are published
- this includes the `javadoc` and `sources` jar files
- we should probably keep only the published artifacts, for now we only have 4 of them. However we could maybe use the build to generate the list of artifacts to include (a bit of research would be needed for that).

v1v: > we often use github actions artifacts as a way to provide a direct download link for the snapshot versions when doing support, so using a .zip file would be better for non *nix users.

I think `tar` is available by default for Windows:
- https://techcommunity.microsoft.com/t5/containers/tar-and-curl-come-to-windows/ba-p/382409

Otherwise, I might need to figure out how to use `zip` to filter files that need to be archived.

> Also, the list of artifacts included is broader than what is necessary:

Would you like us to focus only on the 4 ones we know? If so, I can create the tarball for those files only

JonasKunz: The projects to include can be found via `./gradlew tasks --all | grep publishMavenPublicationToSonatypeRepository` already. This won't print out the jar files names though, we'd need to add a gradle task for that

SylvainJuge: @jackshirazi checked and there is no problem to open `.tar` files on windows, it's event integrated in the GUI as for `.zip` files.

@JonasKunz found a nice way to list explicitly the published artifacts, so we can reuse this to only include those in the `.tar` archive, this will be added in a separate PR to be merged before this one.

JonasKunz: See #226.

SylvainJuge: #226 has been merged, you can use `./gradlew -q printPublishedCodeArtifacts` to the the list of published artifacts in place of the `find` command.

v1v: see https://github.com/elastic/elastic-otel-java/pull/223/commits/746d393c6718ddfc1213eb433c1d8f47be1f1e36

SylvainJuge (approved):

PR #227, opened 2024-04-24
update upstream 2.3.0

SylvainJuge:
- **fix update script**
- **update to 2.3.0**


common/build.gradle.kts
```diff
@@ -23,6 +24,6 @@ dependencies {
     testImplementation(project(":testing-common"))
     testImplementation("io.opentelemetry:opentelemetry-exporter-otlp")
     testImplementation("io.opentelemetry:opentelemetry-exporter-logging")
-    testImplementation("io.opentelemetry:opentelemetry-api-events")
```
SylvainJuge, line 26: [for reviewer] this was moved to an `-incubator` artifact.

common/src/main/java/co/elastic/otel/common/ElasticAttributes.java
```diff
@@ -38,12 +38,4 @@ public interface ElasticAttributes {
 
   AttributeKey<List<String>> PROFILER_STACK_TRACE_IDS =
       AttributeKey.stringArrayKey("elastic.profiler_stack_trace_ids");
-
```
SylvainJuge, line 41: [for reviewer] those attributes are now part of the incubating semconv.

gradle/libs.versions.toml
```diff
@@ -63,6 +72,8 @@ asm = "org.ow2.asm:asm:9.7"
 
 [bundles]
 
+semconv = ["opentelemetrySemconv", "opentelemetrySemconvIncubating"]
```
SylvainJuge, line 75: [for reviewer] this allows to get both the incubating/stable semconv at once, since they are both expected to be used together most of the time.

inferred-spans/build.gradle.kts
```diff
@@ -14,6 +14,7 @@ dependencies {
   implementation(libs.lmax.disruptor)
   implementation(libs.jctools)
   implementation(project(":common"))
+  implementation(libs.bundles.semconv)
```
SylvainJuge, line 17: [for reviewer] this will add dependencies to both the stable and incubating artifacts.

licenses/more-licences.md
```diff
@@ -1,7 +1,7 @@
 
 # agent
 ## Dependency License Report
-
+_2024-04-24 14:54:09 CEST_
```
jackshirazi, line 4: We wanted to generate with no date so that the output was idempotent. Can be fixed later

jackshirazi (approved):

licenses/more-licences.md
SylvainJuge: That's a very good catch, I forgot to run `updateLicensesAndNotice` which does exactly that and only run `generateLicenseReport`. I've fixed it in the last commit.

PR #239, opened 2024-05-02
use upstream span stacktrace

### LINKED ISSUE #172: Contribute upstream: span stacktrace
- [x] suggest contribution in the SIG meeting/discussion: https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/10973
- ~~if no, then nothing changes and we can close this issue for now.~~
- if yes, then next steps are likely:
  - [x] create PR (in java-contrib repo) to contribute upstream, which might also require to also contribute our custom way to modify the spans https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1255
  - [ ] replace the current implementation with the upstream version ~~once it has been released~~ released in contrib 1.35.0 ([release notes](https://github.com/open-telemetry/opentelemetry-java-contrib/releases/tag/v1.35.0)).

Additional challenges:
- [x] we need to exclude the inferred spans, but inferred spans are specific to our distro for now (can we use a transient, not exported, span attribute on inferred spans to avoid this ?) --> Solved by adding an extra heuristic to filter which spans should have the attribute or not.

SylvainJuge:
Depends on [PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1297) to be merged + released to avoid relying on a local snapshot.

Final step of #172 

Fixes #172 

### Checklist

- [x] replace implementation with the upstream one + add minimal tests to ensure it's working as expected.
- [x] replace snapshot once contrib 1.36.0 has been released with upstream span stacktrace artifact
- [x] document usage of span stacktrace module in upstream repo: https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/span-stacktrace


gradle/libs.versions.toml
```diff
@@ -10,11 +10,11 @@ opentelemetryProto = "1.3.1-alpha"
 
 # otel agent, we rely on the '*-alpha' and get the non-alpha dependencies transitively
 # updated from upstream agent with .ci/update-upstream.sh
-opentelemetryJavaagentAlpha = "2.3.0-alpha"
+opentelemetryJavaagentAlpha = "2.4.0-alpha"
 
 # otel contrib
 # updated from upstream agent with .ci/update-upstream.sh
-opentelemetryContribAlpha = "1.34.0-alpha"
+opentelemetryContribAlpha = "1.36.0-alpha"
```
SylvainJuge, line 17: [for reviewer] this is the first version where the contrib artifact is published

gradle/libs.versions.toml
```diff
@@ -10,11 +10,11 @@ opentelemetryProto = "1.3.1-alpha"
 
 # otel agent, we rely on the '*-alpha' and get the non-alpha dependencies transitively
 # updated from upstream agent with .ci/update-upstream.sh
-opentelemetryJavaagentAlpha = "2.3.0-alpha"
+opentelemetryJavaagentAlpha = "2.4.0-alpha"
```
SylvainJuge, line 13: [for reviewer] updating to the latest version allows to also bump the latest SDK and align with the version used for contrib (maybe a further improvement would be to ensure that we don't have any mismatch).

agentextension/build.gradle.kts
```diff
@@ -40,6 +40,8 @@ tasks {
       include("LICENSE")
       include("NOTICE")
       include("licenses/**")
+
+      // TODO : ignore classes that must remain in the agent as we do in 'relocateJavaagentLibs'
```
SylvainJuge, line 44: [for reviewer] follow-up item to create as the extension currently includes classes from API/SDK through transitive dependencies.

buildSrc/src/main/kotlin/elastic-otel.agent-packaging-conventions.gradle.kts
```diff
@@ -81,15 +81,15 @@ tasks {
...
-    }
+
+    exclude("io/opentelemetry/api/")
+    // SDK should not be included, but we have one extra addition to it
+    exclude { it.path.startsWith("io/opentelemetry/sdk") and !it.name.contains("FieldBackedSpanValueStorageProvider.class") }
+    exclude("io/opentelemetry/semconv/")
+    exclude("io/opentelemetry/context/")
+    exclude("io/opentelemetry/internal/")
+    // metrics advice API
+    exclude("io/opentelemetry/extension/incubator/")
```
SylvainJuge, line 92: [for reviewer] For a reason I ignore the exclusion did not work as expected and I had to use file paths to select which classes should be included or not.

This allows to keep `implementation` as dependency scope for dependencies (and not `compileOnly`) and still ensure that we don't accidentally ship API/SDK classes in the bootstrap classloader.

common/build.gradle.kts
```diff
@@ -15,6 +15,9 @@ dependencies {
     compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi")
     implementation(libs.bundles.semconv)
 
+    implementation(libs.contribSpanStacktrace)
```
JonasKunz, line 18: IMO this should not be a implementation (and therefore transitive) depenency of our common project. This way, we pull in the extension into other extensions, which we don't want.

See also my other comment about not replacing our local copy of `MutableSpan`.

Instead, this dependency needs to be compile-only because we still need to support unwrapping the `MutableSpans` if that extension happens to be on the classpath. You'll have to revisit the span unwrapping logic in `SpanValue`.

I'd suggest adding a `static final Map<Class<?>, Function<Span,Span>> UNWRAPPER_REGISTRY` and populating it dynamically based on whether the `MutableSpan` shipped with the span-stacktrace extension is found on the classpath at runtime or not.

JonasKunz (changes requested):

common/build.gradle.kts
JonasKunz: Or, easier solution: Just have a `List<Function<Span,Span>> UNWRAPPERS` populated based on the classpath.
The functions return the unwrapped span (if they know the type and can unwrap) or return the provided span as-is.

The unwrapping logic would then do the following:
```
public Span unwrap(Span toUnwrap) {
    Span previous = null;
    Span current = toUnwrap;
    while (previous != current) {
        previous = current;
        for (Function<Span,Span> unwrapper: UNWRAPPERS) {
             current = unwrapper.apply(current);
        }
    }
    return current;
}
```

custom/build.gradle.kts
```diff
@@ -6,14 +6,15 @@ dependencies {
...
   implementation(project(":universal-profiling-integration"))
-  compileOnly(project(":bootstrap"))
   implementation(project(":resources"))
   compileOnly("io.opentelemetry:opentelemetry-sdk")
   compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi")
   compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
   compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
   compileOnly(libs.bundles.semconv)
 
+  implementation(libs.contribSpanStacktrace)
```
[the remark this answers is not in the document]
SylvainJuge: Excluding transitive dependencies is close to use `compileOnly` instead of `implementation`: we compile against the API, but don't pull transitive dependencies when used.

The downside of this approach is that we then have to pull back needed dependencies again, for example with tests in this module require an extra `testImplementation`, and all the modules that depend on `custom` will have to do the same.

In a sense we need to have the ability to express "use this for implementation, but some transitive dependencies need to be filtered-out in packaging", and I don't know if we can do that with a custom dependency type.

common/build.gradle.kts
SylvainJuge: I think you are right here, we should keep the current MutableSpan and remove it once we have a cleaner solution with the API.

custom/build.gradle.kts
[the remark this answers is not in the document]
JonasKunz: > The downside of this approach is that we then have to pull back needed dependencies again, for example with tests in this module require an extra testImplementation, and all the modules that depend on custom will have to do the same.

My understanding is that `custom` is not intended to be used by anyone but as an aggregator for building our agent-distro and the extension. both of those don't want to have the transitive dependencies.

For having to pull in the transitive dependencies for tests:
Simply adding `testImplementation(libsContribSpanStacktrace)` should pull in everything that has been excluded.
Alternatively, the commonly used OpenTelemetry modules for tests could be added to the test classpath via our `testing-common` module or an convention.

The main downside I see with manually removing the "bad" classes at packaging time is that it makes the build process depend on package names, which are neither public nor guaranteed to be stable (e.g. `io/opentelemetry/extension/incubator/`), making the build potentially prone to errors which are hard to find.

In my opinion the extension (e.g. span-stacktrace) should not be pulling in SDK dependencies via `implementation` anyway and use `compileOnly` , because it doesn't make just our live harder: It also makes the life harder for people who want to package those extensions as their own agent-extension fatJar, becuase they have to do the same exclusions.
For people who use the SDK directly I'd expect them to add the SDK to their classpath explicitly and not rely on transitive dependencies of an SDK-extension, which is kind of the wrong way around.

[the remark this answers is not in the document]
JonasKunz: This might also be something to discuss with the Otel Java SIG to see if they have any opinion on it. Looking at the contrib repo, most extensions there pull in the SDK + API as `API`, which has the downsides (and IMO no upsides) explained above.

universal-profiling-integration/build.gradle.kts
```diff
@@ -24,6 +24,9 @@ dependencies {
   implementation(libs.lmax.disruptor)
   implementation(libs.hdrhistogram) //only used for the WriterReaderPhaser
 
+  compileOnly(libs.contribSpanStacktrace) // for MutableSpan
+  testImplementation(libs.contribSpanStacktrace)
```
JonasKunz, line 28: These dependencies are not needed anymore I think?

agentextension/build.gradle.kts
JonasKunz, line 44: Is this todo still required?

common/src/main/java/co/elastic/otel/common/SpanValue.java
```diff
@@ -271,9 +272,13 @@ private static Span unwrap(Object span) {
     if (span instanceof MutableSpan) {
       return unwrap(((MutableSpan) span).getOriginalSpan());
     }
+    if (CONTRIB_MUTABLE_SPAN_CLASS != null && CONTRIB_MUTABLE_SPAN_CLASS.isInstance(span)) {
```
JonasKunz, line 275: I'd suggest that we add a unit test for this logic:

It should be as easy as just adding a new test arguments [here](https://github.com/elastic/elastic-otel-java/blob/e61b4bd32b2a12457152a052613bde77da59f165/common/src/test/java/co/elastic/otel/common/SpanValueTest.java#L116):
```
Arguments.of(
            Named.of(
                "Contrib MutableSpan, sparse SpanValue",
                ValueAccess.create(
                    SpanValue.<String>createSparse(),
                    io.opentelemetry.contrib.stacktrace.internal.MutableSpan.makeMutable((ReadableSpan) newSpan())))));
```

agentextension/build.gradle.kts
SylvainJuge: Yes, I think so, if you look at the content of the built extension jar it contains the API, SDK and the semantic conventions, I plan to do this in a separate PR though.

JonasKunz: Ah, I see. Thanks for clarifying.

JonasKunz (approved):

agentextension/build.gradle.kts
SylvainJuge: Actually after double-checking this was due to my previous attempt to ignore at file-level, only the semconv attributes are duplicated and that's very probably not a major issue. So I'll remove this comment to prevent any confusion.

PR #249, opened 2024-05-07
Fix ClassCastException in inferred spans

### LINKED ISSUE #236: ClassCastException with inferred spans
```
[otel.javaagent 2024-04-30 14:39:00:847 +0000] [elastic-inferred-spans] WARN co.elastic.otel.profiler.SamplingProfiler - Removing call tree for thread 38 because of exception while handling activation event: class java.lang.ClassCastException class io.opentelemetry.javaagent.shaded.io.opentelemetry.api.trace.PropagatedSpan cannot be cast to class io.opentelemetry.sdk.trace.ReadableSpan (io.opentelemetry.javaagent.shaded.io.opentelemetry.api.trace.PropagatedSpan is in unnamed module of loader 'bootstrap'; io.opentelemetry.sdk.trace.ReadableSpan is in unnamed module of loader io.opentelemetry.javaagent.bootstrap.AgentClassLoader @7ba4f24f)
```
using agent version 0.3.0

JonasKunz:
Fixes #236.

The reason this was not caught by our integration tests is that for the inferred spans test there was no service-name configured. This in turn caused the universal profiling integration to not be active, which triggers the problematic `LocalRootSpan`.

Configuring the `service.name` in the integration test reproduced the issue.

@jackshirazi would still be great if you could do a quick manual test in your scenario where you uncovered the problem to make sure nothing else breaks.

### DIFF
--- a/common/src/main/java/co/elastic/otel/common/LocalRootSpan.java
+++ b/common/src/main/java/co/elastic/otel/common/LocalRootSpan.java
@@ -108,10 +108,12 @@ public class LocalRootSpan {
    */
   @Nullable
   public static ReadableSpan getFor(Span span) {
-    if (span.getSpanContext().isRemote()) {
+    if (span instanceof ReadableSpan) {
+      return getFor((ReadableSpan) span);
+    } else {
+      //This can happen when invoked for a PropagatedSpan (e.g. Span.fromContext())
       return null;
     }
-    return getFor((ReadableSpan) span);
   }
 
   /** See {@link LocalRootSpan#getFor(Span)}. */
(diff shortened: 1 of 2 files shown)


jackshirazi:
Confirmed fixed in the test, inferred spans seen, no errors on profiling

[otel.javaagent 2024-05-07 11:32:00:763 +0000] [main] INFO io.opentelemetry.javaagent.tooling.VersionLogger - opentelemetry-javaagent - version: 0.3.2-SNAPSHOT

image

jackshirazi (approved):

PR #261, opened 2024-05-16
Fix handling of remote spans for inferred spans

### LINKED ISSUE #250: Check the behaviour of inferred spans with remote span activation
OpenTelemetry has the concept of remote spans:
When an incoming W3C-tracecontext is encountered, it is parsed and put as a `PropagatedSpan` (via `Span.fromRemoteContext`) into the current OpenTelemetry context.

This means, `PropagatedSpan`s can be activated and deactivated just like normal spans.
Because we didn't have an equivalent concept in our [elastic-apm-agent](https://github.com/elastic/apm-agent-java/tree/main/elastic-apm-agent), I didn't account for this case when porting the implementation over.
This means that we potentially might end up creating inferred spans with a remote-span as parent, which definitely don't want.

We should add a test case for this and fix the behaviour if required.

JonasKunz:
Closes #250.

Activations of remote spans wil lbe effectively ignored:
 * If a remote span is activated as first span on a thread (typical case), it won't start a profiling session. Only when an actual server-span (child of the remote span) is started, the profiling will be started.
 * If a remote span is activated while a profiling session is already active (=it replace a local span), this activation is effectively ignored. This means inferred spans on the given thread will continue to be children of the previously active span.
 
 @AlexanderWert you observed some strange behaviour of invalid parent/child relationships with a GRPC use-case. Would be great if you find the time to check with the snapshot from this PR if it fixes the issue.

### DIFF
--- a/inferred-spans/src/test/java/co/elastic/otel/profiler/SamplingProfilerTest.java
+++ b/inferred-spans/src/test/java/co/elastic/otel/profiler/SamplingProfilerTest.java
@@ -284,15 +284,15 @@ class SamplingProfilerTest {
     setupProfiler(true);
     awaitProfilerStarted(setup.profiler);
 
-    SpanContext dummyParentCtx = SpanContext.createFromRemoteParent(
-        "a1a2a3a4a5a6a7a8b1b2b3b4b5b6b7b8",
-        "c1c2c3c4c5c6c7c8",
-        TraceFlags.getSampled(),
-        TraceState.getDefault()
-    );
+    SpanContext dummyParentCtx =
+        SpanContext.createFromRemoteParent(
+            "a1a2a3a4a5a6a7a8b1b2b3b4b5b6b7b8",
+            "c1c2c3c4c5c6c7c8",
+            TraceFlags.getSampled(),
+            TraceState.getDefault());
     Span remoteParent = Span.wrap(dummyParentCtx);
     try (Scope scope = remoteParent.makeCurrent()) {
-      //ensure that a remote span activation does not trigger profiling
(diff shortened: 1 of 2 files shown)


inferred-spans/src/test/java/co/elastic/otel/profiler/SamplingProfilerTest.java
```diff
@@ -275,6 +279,49 @@ void testVirtualThreadsExcluded() throws Exception {
...
+    Optional<SpanData> localRoot =
+        spans.stream().filter(s -> s.getName().equals("local-root")).findAny();
+    assertThat(localRoot).isPresent();
+    assertThat(localRoot.get()).hasParentSpanId("c1c2c3c4c5c6c7c8");
+
+    assertThat(spans)
+        .anySatisfy(
+            span -> {
+              assertThat(span).hasParent(localRoot.get());
+              assertThat(span).hasAttribute(ElasticAttributes.IS_INFERRED, true);
```
SylvainJuge, line 321: [minor] could we assert on the inferred span name so it's easier to understand that it's created from the call to `Thread.sleep` ?

SylvainJuge (approved):

inferred-spans/src/test/java/co/elastic/otel/profiler/SamplingProfilerTest.java
JonasKunz: That's ufnortunately not possible without adding significant amounts of, because the inferred spans might or might not create an inferred span for the `testTransactionWithRemoteParent()` method depending on timing.

PR #263, opened 2024-05-23
add docker file

jackshirazi:
Note
- We're only doing the wolfi dockerfile (the image needs `cp` for the otel k8s operator, so we can't start from scratch)
- The /javaagent.jar path is the path the otel k8s operator needs so that the image can be used for auto-instrumentation
- Including the extension jar means the docker image can alternatively be used by the otel k8s operator as an extension for the vanilla otel agent. This can be any path, using an `extensions` dir seems like a reasonable choice

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,6 @@
+FROM docker.elastic.co/wolfi/chainguard-base@sha256:9f940409f96296ef56140bcc4665c204dd499af4c32c96cc00e792558097c3f1
+ARG JAR_FILE
+ARG EXTENSION_JAR_FILE
+COPY ${JAR_FILE} /javaagent.jar
+COPY ${EXTENSION_JAR_FILE} /extensions/elastic-otel-agentextension.jar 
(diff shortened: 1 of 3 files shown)


v1v (approved):
Do you think we can move away from the shell scripts and use the GitHub actions native support?

We did that for some other APM Agents, and it helped us to use the provenance/attes with GitHub actions.

https://github.com/elastic/apm-agent-nodejs/blob/1241a5845a5e4727ee5a833555dafe7e5d2d3034/.github/workflows/release.yml#L47-L77 is how we use it if it helps


In fact, I did some thing for the apm-agent-java but didn't transition to ready:
- https://github.com/elastic/apm-agent-java/pull/3626

Maybe we can apply that new mechanism to this project, what do you think? I'm happy to collaborate on this PR and make that happen if it helps

jackshirazi:
Leaving this unclosed for now, but https://github.com/elastic/elastic-otel-java/pull/268 is the action based implementation

jackshirazi:
Won't use this, we'll use the github actions version instead

PR #268, opened 2024-05-28
Add docker actions

jackshirazi:
I tried to test this, the docker image built but it failed on attestation (maybe because push was false in testing?)


v1v:
> maybe because push was false in testing?)

maybe `push-to-registry: true` was set?

.github/workflows/release.yml
```diff
@@ -128,6 +129,65 @@ jobs:
...
 
+  build-and-push-docker-images:
+    name: "Build and push docker images"
+    runs-on: ubuntu-latest
+    needs:
+      - await-maven-central-artifact
+      - create-github-release
+    env:
+      DOCKER_IMAGE_NAME: docker.elastic.co/observability/elastic-otel-javaagent
+    steps:
```
v1v, line 140:
```suggestion
    permissions:
      attestations: write
      contents: read
      id-token: write
    steps:
```

.github/workflows/release.yml
```diff
@@ -128,6 +129,65 @@ jobs:
...
+          file: "docker/Dockerfile"
+          push: true
+          tags: ${{ steps.docker-meta.outputs.tags }}
+          labels: ${{ steps.docker-meta.outputs.labels }}
+          build-args: |
+            JAR_FILE=elastic-otel-javaagent.jar
+            EXTENSION_JAR_FILE=elastic-otel-agentextension.jar
+
+      - name: generate build provenance (containers)
+        uses: github-early-access/generate-build-provenance@main
```
v1v, line 185:
```suggestion
        actions/attest-build-provenance@173725a1209d09b31f9d30a3890cf2757ebbff0d  # v1.1.2
```

.github/workflows/release.yml
```diff
@@ -128,6 +129,65 @@ jobs:
...
+      - name: Get latest tag
+        run: echo "LATEST_TAG=$(git tag --list --sort=version:refname "v*" | grep -v RC | sed s/^v// | tail -n 1)" >> "${GITHUB_ENV}"
+
+      - uses: elastic/apm-pipeline-library/.github/actions/docker-login@current
+        with:
+          registry: docker.elastic.co
+          secret: secret/apm-team/ci/docker-registry/prod
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
```
v1v, line 155: If possible, let's remove this dependency with Vault and use GitHub secrets. I'll prepare those secrets and then suggest the change here

v1v:
```suggestion
      - name: Log in to the Elastic Container registry
        uses: docker/login-action@e92390c5fb421da1463c202d546fed0ec5c39f20 # v3.1.0
        with:
          registry: ${{ secrets.ELASTIC_DOCKER_REGISTRY }}
          username: ${{ secrets.ELASTIC_DOCKER_USERNAME }}
          password: ${{ secrets.ELASTIC_DOCKER_PASSWORD }}
```

jackshirazi:
> maybe `push-to-registry: true` was set?

it was, you're right

v1v:
> it was, you're right

I think the lack of specific permissions in the job where the attest happened were the reason :)

v1v (approved):

PR #315, opened 2024-07-04
Added portable musl build for jvmti-library

### LINKED ISSUE #228: Jvmti-access lib doesn't support Alpine Linux
Currently the universal-profiling-integration crashes on linux distros shipping musl instead of glibc (e.g. Alpine):
```
[otel.javaagent 2024-04-25 08:31:34:144 +0000] [main] DEBUG co.elastic.otel.UniversalProfilingProcessor - Opening profiler correlation socket /tmp/essockGld1x67O
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000000003250, pid=1, tid=7
#
# JRE version: OpenJDK Runtime Environment Temurin-21.0.3+9 (21.0.3+9) (build 21.0.3+9-LTS)
# Java VM: OpenJDK 64-Bit Server VM Temurin-21.0.3+9 (21.0.3+9-LTS, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-aarch64)
# Problematic frame:
# C  [elastic-jvmti-linux-arm64-4813494d137e1631bba301d5acab6e7b-b804a293859bee2eb1968b3ee9e92141.so+0x3890]  init_have_lse_atomics+0xc
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# //hs_err_pid1.log
#
# If you would like to submit a bug report, please visit:
#   https://github.com/adoptium/adoptium-support/issues
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#

```

This can only be prevented at the moment by disabling the profiling integration via `ELASTIC_OTEL_UNIVERSAL_PROFILING_INTEGRATION_ENABLED=false`.

We should fix this by either:
 * statically linking against musl instead of dynamically linking against glibc
 * providing a separate musl binary for linux (x64 and arm64).

JonasKunz:
Fixes #228 , but needs some more investigation:
It looks like the UP integration is partly broken on musl. The profiling data will contain the service name, but the trace/span-id correlation seems to not work.

### DIFF
--- a/jvmti-access/src/main/jni/ProfilerSocket.cpp
+++ b/jvmti-access/src/main/jni/ProfilerSocket.cpp
@@ -3,6 +3,7 @@
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <fcntl.h>
+#include <cstring>
 
 namespace elastic {
     namespace jvmti_agent {
--- a/jvmti-access/build.gradle.kts
+++ b/jvmti-access/build.gradle.kts
@@ -74,6 +74,16 @@ val nativeTargets = listOf(
     "linux-x64.so",
     "jni_linux_x64.Dockerfile",
     "-static-libstdc++ -static-libgcc -mtls-dialect=gnu2 $sharedCompilerArgs"
+  ),
+  NativeTarget(
+    "linux-musl-arm64.so",
+    "jni_linux_musl_arm64.Dockerfile",
+    "-static-libstdc++ -static-libgcc -mtls-dialect=desc $sharedCompilerArgs"
+  ),
+  NativeTarget(
+    "linux-musl-x64.so",
+    "jni_linux_musl_x64.Dockerfile",
+    "-static-libstdc++ -static-libgcc -mtls-dialect=gnu2 $sharedCompilerArgs"
   )
 )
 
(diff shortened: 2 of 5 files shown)


jvmti-access/src/main/jni/ProfilerSocket.cpp
```diff
@@ -3,6 +3,7 @@
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <fcntl.h>
+#include <cstring>
```
SylvainJuge, line 6: is it required to add it ? does not seem used unless there is something implicit.

JonasKunz: Yup, it is required. Looks like with glibc it was included transitively form one of the other headers, but not for musl

SylvainJuge (approved):
With the 0.3.2 agent I used to get the following error when starting the agent with `eclipse-temurin:17-alpine` Docker image to run the agent with [`cpu-burner` app](https://github.com/JonasKunz/cpu-burner/tree/musl-demo).

```
otel.javaagent 2024-07-10 15:16:10:988 +0000] [main] ERROR co.elastic.otel.JvmtiAccess - Failed to load jvmti native library
java.lang.UnsatisfiedLinkError: /tmp/elastic-jvmti-linux-x64-4813494d137e1631bba301d5acab6e7b-6ca4b5b2f9e60ed3ef1c59643b609af0.so: Error loading shared library ld-linux-x86-64.so.2: No such file or directory (needed by /tmp/elastic-jvmti-linux-x64-4813494d137e1631bba301d5acab6e7b-6ca4b5b2f9e60ed3ef1c59643b609af0.so)
        at java.base/jdk.internal.loader.NativeLibraries.load(Native Method)
        at java.base/jdk.internal.loader.NativeLibraries$NativeLibraryImpl.open(NativeLibraries.java:388)

```

Re-building this PR locally and using this updated agent, the agent now starts as expected.

PR #322, opened 2024-07-09
Force monotonic clock when starting async-profiler

JonasKunz:
We had a user report that the `inferred-spans` feature fails to generate spans even though it should.
After analyzing the diagnostic files, turns out that the profiling data was using a different clock than `System.nanoTime`, which is required for correlation.

Async-profiler 3.0 by default tries to use `TSC` only falls back to `System.nanoTime` if `TSC` is not available on the current CPU/System. The `clock=m` startup argument should force the monotonic clock (=`System.nanoTime`) to be always used.


JonasKunz:
User confirmed that this change fixes the problem.

SylvainJuge (approved):

PR #326, opened 2024-07-12
Add support for virtual threads to universal profiling integration

### LINKED ISSUE #81: Support virtual threads in profiling integration
#43 added the capability to provide the trace context to the universal profiling host agent via native `thread_local` variables.
This currently does not work for `virtual-threads`: When a virtual thread is unmounted or mounted, the native `thread_local` would get out of sync.

The goal of this issue is to add support for virtual threads by leveraging the JVMTI virtual thread mount/unmount events to automatically keep the native `thread_local` in sync.

JonasKunz:
Closes #81. Hooks on to JVMTI mount / unmount events to keep the trace/span context for the profiler in sync.
Can be disabled via a config option in case anything goes wrong.


jvmti-access/src/main/java/co/elastic/otel/UniversalProfilingCorrelation.java
```diff
@@ -184,8 +198,7 @@ private static MethodHandle generateVirtualChecker() {
       return MethodHandles.lookup().unreflect(isVirtual);
     } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
       // virtual threads are not supported, therefore no thread is virtual
-      return MethodHandles.dropArguments(
-          MethodHandles.constant(boolean.class, false), 0, Thread.class);
+      return null;
```
SylvainJuge, line 201: [question] why do we don't need to call `dropArguments` anymore here ? Was it not necessary in the first place ?

jvmti-access/src/main/jni/VirtualThreadSupport.cpp
```diff
@@ -0,0 +1,214 @@
+#include "VirtualThreadSupport.h"
+#include <cstring>
+
+namespace elastic {
+    namespace jvmti_agent {
+
+        namespace {
+
+            bool isExpectedMountEvent(jvmtiExtensionEventInfo& eventInfo) {
```
SylvainJuge, line 9: [minor] those two methods seem very close or identical, would there be benefits to merging them in a single one ?

SylvainJuge (approved):
LGTM, however I'll trust the tests more than my understanding of the C++ part :-).

jvmti-access/src/main/java/co/elastic/otel/UniversalProfilingCorrelation.java
JonasKunz: I changed the constant to be `null` instead of being a `MethodHandle` always returning `false`.
I did this to have an easy check for whether virtual threads are available on the platform or not (comparing the constant against null).

jvmti-access/src/main/jni/VirtualThreadSupport.cpp
JonasKunz: I think it's a case of incidental duplication: This methods check for the signature of two different events (unmount and mount) which are not guaranteed to evolve together, so it is better to keep the duplication here imo.

PR #344, opened 2024-08-08
Add update-cli task to automatically update upstream agent

JonasKunz:
Automatically opens a PR to update to the latest upstream agent.
Uses the existing `gradle/update-upstream.sh` to also correctly update dependent libraries (semconv and contrib).

### DIFF
--- a/.ci/updatecli.d/bump-upstream-agent-version.yml
+++ b/.ci/updatecli.d/bump-upstream-agent-version.yml
@@ -8,7 +8,7 @@ scms:
     kind: github
     spec:
       user: '{{ requiredEnv "GITHUB_ACTOR" }}'
-      owner: JonasKunz
+      owner: elastic
       repository: elastic-otel-java
       token: '{{ requiredEnv "GITHUB_TOKEN" }}'
       username: '{{ requiredEnv "GITHUB_ACTOR" }}'
--- a/.ci/updatecli.d/bump-upstream-agent-version.yml
+++ b/.ci/updatecli.d/bump-upstream-agent-version.yml
@@ -51,7 +51,7 @@ targets:
     kind: shell # toml action currently erases comments, so we fallback to regex-replace
     scmid: githubConfig
     spec:
-      command: gradle/update-upstream.sh v{{ source "upstream-agent-version" }} && git add . & git commit -m "Update upstream versions"
+      command: gradle/update-upstream.sh v{{ source "upstream-agent-version" }} && echo "Update successful"
 
 actions:
   open-pr:
--- a/.ci/updatecli.d/bump-upstream-agent-version.yml
+++ b/.ci/updatecli.d/bump-upstream-agent-version.yml
@@ -8,7 +8,7 @@ scms:
     kind: github
     spec:
       user: '{{ requiredEnv "GITHUB_ACTOR" }}'
-      owner: elastic
+      owner: JonasKunz
       repository: elastic-otel-java
       token: '{{ requiredEnv "GITHUB_TOKEN" }}'
       username: '{{ requiredEnv "GITHUB_ACTOR" }}'
--- a/.ci/updatecli.d/bump-upstream-agent-version.yml
+++ b/.ci/updatecli.d/bump-upstream-agent-version.yml
@@ -46,7 +46,6 @@ targets:
 
   update-agent-version:
     name: "Update upstream OpenTelemetry agent version and related dependencies"
-    conditionids: [update-required]
     disablesourceinput: true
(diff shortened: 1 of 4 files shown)


.github/workflows/updatecli.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: elastic/oblt-actions/updatecli/run@v1
+        with:
+          command: "--experimental apply --config .ci/updatecli.d"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - if: failure()
+        uses: elastic/oblt-actions/slack/send@v1
+        with:
+          bot-token: ${{ secrets.SLACK_BOT_TOKEN }}
```
JonasKunz, line 26: Does the slackbot token exist on this repo?

.github/workflows/updatecli.yml
```diff
@@ -0,0 +1,28 @@
+name: updatecli
+
+on:
+  workflow_dispatch: ~
+  schedule:
+    - cron: '0 6 * * *'
+
+permissions:
+  contents: write
```
JonasKunz, line 9: Is using the standard GH actions bot with these permissions okay here?

.github/workflows/updatecli.yml
v1v: yes

.ci/updatecli.d/bump-upstream-agent-version.yml
```diff
@@ -0,0 +1,66 @@
...
+targets:
+
+  update-agent-version:
+    name: "Update upstream OpenTelemetry agent version and related dependencies"
+    conditionids: [update-required]
+    disablesourceinput: true
+    kind: shell # toml action currently erases comments, so we fallback to regex-replace
+    scmid: githubConfig
+    spec:
+      command: gradle/update-upstream.sh v{{ source "upstream-agent-version" }} && echo "Update successful"
```
v1v, line 54: What's the reason for `echo "Update successful"`?

.ci/updatecli.d/bump-upstream-agent-version.yml
```diff
@@ -0,0 +1,66 @@
...
+    spec:
+      command: test "{{ source "upstream-agent-version" }}-alpha" != "{{ source "distro-otel-agent-version" }}"
+
+targets:
+
+  update-agent-version:
+    name: "Update upstream OpenTelemetry agent version and related dependencies"
+    conditionids: [update-required]
+    disablesourceinput: true
+    kind: shell # toml action currently erases comments, so we fallback to regex-replace
```
v1v, line 51: required comment?

.ci/updatecli.d/bump-upstream-agent-version.yml
```diff
@@ -0,0 +1,66 @@
...
+    disablesourceinput: true
+    kind: shell
+    spec:
+      command: test "{{ source "upstream-agent-version" }}-alpha" != "{{ source "distro-otel-agent-version" }}"
+
+targets:
+
+  update-agent-version:
+    name: "Update upstream OpenTelemetry agent version and related dependencies"
+    conditionids: [update-required]
```
v1v, line 49: What's the reason for this condition? IIUC, conditions will be the ones deciding whether the targets can be processed or no.

.github/workflows/updatecli.yml
v1v: This won't trigger a build. In this particular case we use

```suggestion
  contents: read
```


And a new secret called `UPDATECLI_GH_TOKEN`. I'll need to create the secret

.ci/updatecli.d/bump-upstream-agent-version.yml
JonasKunz: We only want to perform an update if the current version isn't up to date.
That's why I though we would need the condition, don't we?

v1v: That's the default behaviour, you don't need to be explicit as far as I recall.

For instance:
- https://github.com/elastic/opbeans-java/blob/main/.ci/updatecli.d/bump-version.yml#L44-L59

.ci/updatecli.d/bump-upstream-agent-version.yml
JonasKunz: Yes, but outdated. The [toml](https://www.updatecli.io/docs/plugins/resource/toml/) target would be perfect here, but we can't use it because it reformats the toml file and erases comments

v1v (approved):

PR #366, opened 2024-08-13
Replace groovy with kotlin for root project

JonasKunz:
Finally cleans up the root project by replacing groovy with kotlin.

Also, it is discouraged to add configuration via `subProjects` because it makes the source of the effective configuration hard to spot. These configurations have been moved into the exisitng `java-conventions` and the new `spotless-conventions` projects instead.

I've manually tested the `setVersion`, `setNextVersion` and `currentVersion` tasks. I'll have an eye on the snapshot-publishing after this PR is merged to ensure that publishing still works correctly.


build.gradle
```diff
@@ -1,127 +0,0 @@
...
-    versionFile.text = versionFile.text.replaceFirst(pattern) { match, version, suffix ->
-      def (major, minor, patch) = version.split(/\./).collect { it as int }
-      "version=${major}.${minor}.${patch + 1}-SNAPSHOT"
-    }
-  }
-}
-
-tasks.configure {
-  doLast {
-    copyGitHooks
```
JonasKunz, line 125: I did not migrate the automatic installation of git hooks over: I think this was broken before, because when I now had them installed, comitting from inside the IDE would break. So I think it is best to have them in the repo and have the installation be opt-in

SylvainJuge: If it's broken for you then it's better to keep it opt-in, it's just a convenience anyway as the build fails if the formatting is not correct.

buildSrc/src/main/kotlin/elastic-otel.java-conventions.gradle.kts
```diff
@@ -1,12 +1,50 @@
 plugins {
-  id("java")
+  java
+  id("elastic-otel.spotless-conventions")
```
SylvainJuge, line 3: Why do we need to have spotless conventions in a separate file ? Is it to allow applying it both for the conventions themselves in addition to the project modules ?

SylvainJuge (approved):

buildSrc/src/main/kotlin/elastic-otel.java-conventions.gradle.kts
JonasKunz: We could use spotless for non-java projects, though I don't think we'd ever do that. I did this more for a separation of concerns. If you prefer to have it directly in the java-conventions, I can put it in. I don't have a strong opinion here.

SylvainJuge: Me neither, let's keep it this way then as it's usually easier to merge than to split later :-).

PR #369, opened 2024-08-14
span stacktraces config (without aligning with upstream)

SylvainJuge:
Relates to https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1414 in the upstream contrib repo, constant is duplicated and needs to be kept in sync until available in a new contrib release.

Update: using `elastic.otel.java.span-stacktrace.min.duration` as the upstream config is not yet stable and will require further discussion, so let's stick to elastic namespace for now.


SylvainJuge:
Currently blocked on having the https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1414 reviewed and merged, as until it's merged the config option naming is not set in stone.

jackshirazi:
and docs/configure.md

jackshirazi (approved):

PR #370, opened 2024-08-15
Replace inferred-spans extension with upstream contrib extension

### LINKED ISSUE #174: Contribute upstream: inferred spans
depends on the update to async-profiler 3.x update: #173

JonasKunz:
Closes #174.
Replaces the inferred-spans extension with the contributed upstream one, but maintains a backwards compatibility layer for the `ELASTIC_` options because they are used a lot already in demos.

This PR is also kind of blocked by https://github.com/elastic/apm-data/pull/321. That change did not make it into `8.15`. We'll need that change to ship, otherwise the parent-child overriding won't wokr correctly, because the attribute `elastic.is_child` was renamed to `is_child`.


JonasKunz:
Support for `is_child` without the elastic prefix will be part of the 8.15.1 release.

JonasKunz:
https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1533 is now part of the included contrib version, so this PR is ready for review

common/src/main/java/co/elastic/otel/common/ElasticAttributes.java
```diff
@@ -30,11 +30,11 @@ public interface ElasticAttributes {
...
-  /** Marker attribute for inferred spans. */
-  AttributeKey<Boolean> IS_INFERRED = AttributeKey.booleanKey("elastic.is_inferred");
-
-  /** Used as marker on span-links to override the parent-child relationship for inferred spans. */
-  AttributeKey<Boolean> IS_CHILD = AttributeKey.booleanKey("elastic.is_child");
+  /**
+   * Marker attribute for inferred spans. Does not have the elastic-prefix anymore because it has
+   * been contributed upstream
+   */
+  AttributeKey<Boolean> IS_INFERRED = AttributeKey.booleanKey("is_inferred");
```
SylvainJuge, line 37: [minor] do we have a way to reuse the upstream attribute without duplicating it ?

inferred-spans/src/test/java/co/elastic/otel/InferredSpansBackwardsCompatibilityConfigTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+  private SamplingProfiler extractProfiler(InferredSpansProcessor processor) {
+    try {
+      Field profilerField = processor.getClass().getDeclaredField("profiler");
+      profilerField.setAccessible(true);
+      return (SamplingProfiler) profilerField.get(processor);
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
```
SylvainJuge, line 105: [minor] wouldn't it be better to avoid using reflection and rely on a package-private helper class here ?

inferred-spans/src/test/java/co/elastic/otel/InferredSpansBackwardsCompatibilityConfigTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+  @Test
+  @DisabledOnOpenJ9
+  public void checkAllLegacyOptions(@TempDir Path tmpDir) {
+    String libDir = tmpDir.resolve("foo").resolve("bar").toString();
+    try (AutoConfigTestProperties props =
+        new AutoConfigTestProperties()
+            .put("elastic.otel.inferred.spans.enabled", "true")
+            .put("elastic.otel.inferred.spans.logging.enabled", "false")
+            .put("elastic.otel.inferred.spans.backup.diagnostic.files", "true")
```
SylvainJuge, line 61: maybe it would be worth adding a test to test the priority between legacy and upstream config options, focusing only on a few basic ones (and we can assume the same priority is applied to others).

SylvainJuge (approved):
Good job and very nice diff karma !

jackshirazi:
are we still going to be generating an extension jar that will be included in the docker image?

common/src/main/java/co/elastic/otel/common/ElasticAttributes.java
JonasKunz: Not here, because we don't want this package to pull in the inferred spans extension as a dependency.

inferred-spans/src/test/java/co/elastic/otel/InferredSpansBackwardsCompatibilityConfigTest.java
JonasKunz: Added in [38532](https://github.com/elastic/elastic-otel-java/pull/370/commits/38532ee1b4dd0b59bee631e8e0aa75772d83e8f7)

jackshirazi:
> However, we'll still be publishing it as part of the [agentextension](https://github.com/elastic/elastic-otel-java/tree/main/agentextension) jar which allows you to load all our distro features into a vanilla agent. But I don't think we put this jar in our docker image, do we?

Yes, we do. The docker image is set up for the otel operator so that it can be used both as an image for running the agent and separately as an image to include our extensions. The latter requires the agentextension jar to be in a [specific location](https://github.com/elastic/elastic-otel-java/blob/54fd86659a02d9310c99bce7c75e2bd779c329ca/docker/Dockerfile#L5).

So your answer is ultimately yes :-)

jackshirazi (approved):

PR #155, opened 2024-02-23
update licences

SylvainJuge:
- **licences update**
- **use version catalog**
- **generate a single file notice file**
- **embed asm bsd 3 clause license**
- **embed MPL 2 license**
- **include NOTICE and LICENSE in libraries (jar) and agent (jar)**

re-opens closed PR due to making the project public: #128 

## Implementation details:

In addition of the project License (Apache2), two extra licenses are included to cover dependencies (they are linked from the `/NOTICE` file)
- `/licenses/LICENSE_asm_bsd-3-clause` 
- `/licenses/LICENSE_mpl-2`

In source code:

- the `/LICENSE` contains a copy of the Apache 2 license
- the `/NOTICE` is generated from a template
- the `/licenses/` folder contains:
  - one `more-licenses.md` file that contains the list of dependencies and their licences with links to sub-folders
  - one sub-folder per embedded dependency of the project with license and notice if they are packaged in the dependency
  
In packaged artifacts (in particular for the java agent):

- the `/LICENSE` file is copied to `/META-INF/LICENSE`
- the `/NOTICE` file is copied to `/META-INF/NOTICE`
- the `/licenses` folder from the upstream agent is copied as-is to `/META-INF/licenses`
- the `/licenses` folder content is copied to `/META-INF/licenses`, hence it complements what is included in the upstream agent.
- two markdown files describe the dependencies `/META-INF/licenses/more-licenses.md` (from this project) and `/META-INF/licenses/licenses.md` (from the upstream agent), some overlap is expected but there is no harm for duplication here.

To update:

- the `update-license-and-notice` gradle task updates the `/NOTICE` and `/licenses/more-licenses.md` + sub-folders.

PR #385, opened 2024-08-21
trigger license update with renovate

SylvainJuge:
Configure renovate to use `postUpgradeTasks` in order to automatically update license files when there are dependency updates.

This prevents obsolete license and PRs like #155, #281 and #381 

Requires to configure `allowedPostUpgradeCommands` in the global configuration.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -18,5 +18,9 @@
     ".github/**",
     ".buildkite/**",
     "jvmti-access/jni-build/*.Dockerfile"
-  ]
+  ],
+  "postUpgradeTasks": {
+    "commands": ["./gradlew updateLicensesAndNotice"],
+    "fileFilters": ["licences/**/*"]
+  }
 }


jackshirazi:
> Requires to configure allowedPostUpgradeCommands in the global configuration.

has that been achieved?

SylvainJuge:
> > Requires to configure allowedPostUpgradeCommands in the global configuration.
> 
> has that been achieved?

No, and this not doable for now, so we will need another approach here.
I need to open a ticket as a hope it will be available within a reasonable future.

I will open a ticket to properly describe the issue and we can think about an other implementation, for example by running the licence update script and check for non-commited changes directly in PRs.

SylvainJuge:
Here is the issue https://github.com/elastic/elastic-otel-java/issues/386

SylvainJuge:
Closing this issue for now as we can't do that with renovate at the moment.

Issue #386, opened 2024-08-23
Automate licenses update


SylvainJuge:
For now we update the project licences by executing `./gradlew updateLicensesAndNotice` manually.

The risk is to forget about it and get obsolete licences definitions, which also means we have to open dummy PRs like #155, #281 and #381 

For now we can't use renovate `postUpgradeTasks` feature as it requires a global configuration (see #385 ) for how this could have helped apply this for dependency updates.

Few possible implementation ideas:
- add a check in build that runs `./gradlew updateLicensesAndNotice` then checks for un-commited changes in the repo with a `git` command (this won't work for dependency update PRs though)
- schedule a GH action that will automatically open a PR when an update is needed
- include those changes in the first release PR before the release artifact is created, that ensures that we update it at least whenever we release.

v1v:
What's stopping us from using GitHub actions to generate the licenses while they discuss solving this with Renovate?

https://github.com/elastic/elastic-otel-java/pull/390/files might help. It's not yet fully finished, but it took me a couple of minutes to configure what we have done in other places and filter by branch and folders.

v1v:
> I think it's a very good work-around for now, will it be included in every PR or is that something that we'll have to schedule on main ?

In every PR

and we could even run on `main` too if that helps, so it can create a PR back...

As you prefer

I'll work on pushing the changes and then we will validate how it goes

JonasKunz:
Taking a step back: Do we actually need the copies of dependency licenses in our repo?
I think we do need them in the agent artifact (because there we literally include the dependencies in the jar), but that could then be simply part of the jar-task instead of having all licensesadded to the git repo?

SylvainJuge:
I don't recall the exact terminology, but when making this project compliant with the OSS policy it was one of the requirements to have it part of the binaries that we ship.

For the source code, we also have to include the notices for those dependencies, and having them checked-in allows to double check when a license changes or when dependencies are added.

I'll link to the internal issue for the OSS publication here.

JonasKunz:
Summary from discussion on 2024/09/02: 
We'll stop checking in the licenses in the git repo all together and will instead package them on the fly when the jar is built via gradle.
This task should then contain a check for approved licenses and fail the build if the license is not known / compliant.

I'll implement this when I'm on duty.

PR #390, opened 2024-08-29
ci(update-license): for renovate PRs

v1v:
#### What

Update the licenses for the Renovate PRs using the gradle task.

Part of https://github.com/elastic/elastic-otel-java/issues/386

I partially took some bits and pieces from https://github.com/elastic/apm-server/blob/main/.github/workflows/update-dependabot-pr.yml?rgh-link-date=2024-08-29T14%3A33%3A13Z

#### Why

Temporary workaround to remove the burden for now when renovate creates PRs, but the licenses need to be updated too.

#### Test
I tested a similar workflow and https://github.com/elastic/observability-robots-playground/pull/55 is the outcome (only accessible for Elastic Employees)

### DIFF
--- a/.github/workflows/update-renovate-pr.yml
+++ b/.github/workflows/update-renovate-pr.yml
@@ -1,0 +1,48 @@
+---
+# Renovate Pull Requests are not updating the license files in the licenses folder.
+# This workflow will checkout the renovate PR, update the license files, and push the changes back to the PR.
+name: update-renovate-pr
+
+on:
+  push:
+    branches:
+      - renovate/**
+    paths-ignore:
+      - licenses/**
+
+permissions:
+  contents: read
+
+jobs:
+  update-renovate-pr:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/workflows/gradle-goal
+        with:
+          command: "./gradlew updateLicensesAndNotice"
+
+      - name: Debug diff
+        run: git status --porcelain
+
+      - name: Check if NOTICE.txt has changed
+        id: check-notice-diff
+        run: |
+          changed=false
+          if git status --porcelain | grep -q 'NOTICE.txt'; then
+            changed=true
+          fi
+          echo "changed=$changed" >> "$GITHUB_OUTPUT"
+
+      - name: Debug output
+        run: |-
+          echo 'changed: ${{ steps.check-notice-diff.outputs.changed }}'
+
+      - name: Commit changes (TBD)
+        if: ${{ steps.check-notice-diff.outputs.changed == 'true' }}
+        run: |
+           git status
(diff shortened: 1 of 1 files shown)


.github/workflows/update-renovate-pr.yml
```diff
@@ -0,0 +1,48 @@
...
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/workflows/gradle-goal
+        with:
+          command: "./gradlew updateLicensesAndNotice"
+
+      - name: Debug diff
+        run: git status --porcelain
+
+      - name: Check if NOTICE.txt has changed
```
SylvainJuge, line 31: we need to check if there is any change in the whole `licenses` folder here, there is no `NOTICE.txt`, the `NOTICE` at the project root is maintained manually.

v1v: C&P from https://github.com/elastic/apm-server/blob/main/.github/workflows/update-dependabot-pr.yml :)

v1v: https://github.com/elastic/elastic-otel-java/pull/390/commits/faf9b581abe3bb6e6eae1f04dc936b6b7ad3847e

.github/workflows/update-renovate-pr.yml
```diff
@@ -0,0 +1,48 @@
...
+          if git status --porcelain | grep -q 'licenses/'; then
+            changed=true
+          fi
+          echo "changed=$changed" >> "$GITHUB_OUTPUT"
+
+      - name: Debug output
+        run: |-
+          echo 'changed: ${{ steps.check-notice-diff.outputs.changed }}'
+
+      - name: Commit changes (TBD)
```
v1v, line 44: We probably need to use our GitHub machine account to ensure the CI runs afterwards

v1v: That's not needed when using https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#create-or-update-file-contents

.github/workflows/update-renovate-pr.yml
```diff
@@ -0,0 +1,68 @@
...
+            changed=true
+          fi
+          echo "changed=$changed" >> "$GITHUB_OUTPUT"
+
+      - name: Debug output
+        run: |-
+          echo 'changed: ${{ steps.check-notice-diff.outputs.changed }}'
+
+      ## this should help with of multiple files in a new folder.
+      - name: Add changes
```
v1v, line 45: This is handy so we can run the push changes through all the files, otherwise all the files under a new folder won't be listed when using `git status --porcelain`

reakaleek (approved):

v1v:
IIUC, this won't be needed, as stated in https://github.com/elastic/elastic-otel-java/issues/386#issuecomment-2324195703. I'll close this now

PR #409, opened 2024-09-19
Generate license report on jar build

### LINKED ISSUE #386: Automate licenses update

JonasKunz:
Closes #386.

Instead of checking licenses into git, the license report is instead generated when the JAR-artifacts are built.
I manually compared the results of the new build with the contents prior to this PR. Prior to this PR the license report also included compile-only dependencies, which are now left out because to my knowledge they are not required: We only need to have the licenses of the dependencies we actually ship.

To manage license approval, I've included the `checkLicense` task: This way the build will fail if licenses are detected, which have not been approved explicitly via the `buildscripts/allowed-licenses.json` file.


agent/build.gradle.kts
```diff
@@ -26,104 +26,30 @@ dependencies {
...
 
   // We override the agent entrypoints defined in elastic-otel.agent-packaging-convention
   shadowJar {
-
-    dependsOn(named("updateLicensesAndNotice"))
-
     // include licenses and notices in jar
-    from(rootDir) {
+    dependsOn(fullLicenseReport)
+    from(fullLicenseReport.get().outputs.files.singleFile) {
```
LikeTheSalad, line 45: Usually passing a task as input for another task is enough for gradle to create an implicit dependency between them and also to know what outputs to consume. Some cases might need things to be done explicitly though, maybe this is one of them, but if it's not, I think it'd be enough to replace lines 44 and 45 with `from(fullLicenseReport)`.

LikeTheSalad: Also, it seems like this config is applied the same way in the `agentextension` build file, if the same steps are needed for all cases then it'd probably be nice to make this work in the conventions file.

LikeTheSalad (approved):

agent/build.gradle.kts
JonasKunz: > I think it'd be enough to replace lines 44 and 45 with from(fullLicenseReport)

Good point, i wasn't aware that `from()` accepts tasks as input, thanks!

> Also, it seems like this config is applied the same way in the agentextension build file, if the same steps are needed for all cases then it'd probably be nice to do this work in the conventions file.

I left that intentionally in the individual sub-projects: In my opinion it is easier to follow when the content-packaging of the JAR is less spread over multiple files, especially in this case because it is now only two duplicated lines across the two files.

agent/build.gradle.kts
```diff
@@ -26,104 +26,27 @@ dependencies {
...
-          "A copy of the HdrHistogram license is provided in the 'licenses/LICENSE_hdrhistogram_bsd-2-clause_cc0' file.",
-          "",
-          "Details on the individual dependencies notices and licenses can be found in the 'licenses' folder.",
-          ""
-        )
-      )
-
-      Files.write(rootDir.toPath().resolve("NOTICE"), lines)
-
-      // make the generated license report idempotent by removing the date
```
SylvainJuge, line 119: This part is still needed otherwise it makes the build non-idempotent due to the date generated in `more-licenses.md`.

licenses/LICENSE_asm_bsd-3-clause
```diff
@@ -1,28 +0,0 @@
-
- ASM: a very small and fast Java bytecode manipulation framework
```
SylvainJuge, line 2: this file had to be added explicitly as a copy of the licence had to be provided, and it's not added in the build-time version.

licenses/LICENSE_hdrhistogram_bsd-2-clause_cc0
```diff
@@ -1,41 +0,0 @@
-The code in this repository code was Written by Gil Tene, Michael Barker,
```
SylvainJuge, line 1: this file had to be added explicitly as a copy of the licence had to be provided, and it's not added in the build-time version.

licenses/LICENSE_mpl-2
```diff
@@ -1,373 +0,0 @@
-Mozilla Public License Version 2.0
```
SylvainJuge, line 1: this file had to be added explicitly as a copy of the licence had to be provided, and it's not added in the build-time version.

agent/build.gradle.kts
```diff
@@ -26,104 +26,27 @@ dependencies {
   upstreamAgent("io.opentelemetry.javaagent:opentelemetry-javaagent")
 }
 
+licenseReport {
+  configurations = arrayOf(
+    project.configurations.bootstrapLibs.name,
+    project.configurations.javaagentLibs.name
+    // No need to include the upstreamAgent, because it already has all dependent licenses packaged
+    // Those will be preserved in our JAR
```
SylvainJuge, line 34: Is is due to this change that we see so many differences in the `more-licences.md` file ? However the `licenses.md` remains perfectly identical with this PR.

buildSrc/src/main/kotlin/elastic-otel.license-report-conventions.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+    exclude("project-licenses-for-check-license-task.json")
+    into("licenses")
+  }
+  into(layout.buildDirectory.dir("reports/all-licenses"))
+}
+
+licenseReport {
+  allowedLicensesFile = File("${rootProject.rootDir}/buildscripts/allowed-licenses.json")
+
+  renderers = arrayOf<ReportRenderer>(InventoryMarkdownReportRenderer("more-licences.md"))
```
SylvainJuge, line 31: [minor] maybe we could align and use `more-licenses.md` to be consistent with `allowed-licenses.json`, the original typo was probably mine :-).

agent/build.gradle.kts
JonasKunz: Done in https://github.com/elastic/elastic-otel-java/pull/409/commits/acda4aced88c62fab8629391ad53d9638283616a

licenses/LICENSE_asm_bsd-3-clause
JonasKunz: We don't provide ASM, is is already included in the upstream agent. I think this was initially added because we use it as a compile-time dependency, for which don't need to include the licenses though because we don't ship them.

licenses/LICENSE_hdrhistogram_bsd-2-clause_cc0
JonasKunz: The license is there:

> **9** **Group:** `org.hdrhistogram` **Name:** `HdrHistogram` **Version:** `2.2.2` 
> - **Manifest License**: The 2-Clause BSD License (Not Packaged)
> - **POM Project URL**: [http://hdrhistogram.github.io/HdrHistogram/](http://hdrhistogram.github.io/HdrHistogram/)
> - **POM License**: Creative Commons Legal Code - [https://creativecommons.org/publicdomain/zero/1.0/legalcode](https://creativecommons.org/publicdomain/zero/1.0/legalcode)
> - **POM License**: PUBLIC DOMAIN - [http://creativecommons.org/publicdomain/zero/1.0/](http://creativecommons.org/publicdomain/zero/1.0/)
> - **POM License**: The 2-Clause BSD License - [https://opensource.org/licenses/BSD-2-Clause](https://opensource.org/licenses/BSD-2-Clause)
> - **Embedded license files**: [HdrHistogram-2.2.2.jar/META-INF/LICENSE.txt](HdrHistogram-2.2.2.jar/META-INF/LICENSE.txt)

licenses/LICENSE_mpl-2
JonasKunz: This license was there for some reason manually for OkHttp, but we don't use it anywhere. For that reason we don't need that license. If we at any time add OkHttp as dependency, it would be added back automatically.

SylvainJuge (approved):

PR #422, opened 2024-10-01
add dynamically disabled instrumentation capability

jackshirazi:
The trace scoping in the SDK is still experimental at this point, so the relevant methods are non-public, hence the use of reflection to access them. Also, dynamically changing scope enablement is additionally experimental with the only additional support in SdkTracer.tracerEnabled being mutable (non-final). As it turns out, this is all that is needed to enable dynamically disabling instrumentation by scope, using reflection.

### DIFF
--- a/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
+++ b/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
@@ -204,13 +204,20 @@ public class DynamicInstrumentation {
   static {
     if ("true".equals(System.getProperty(INSTRUMENTATION_DISABLE_OPTION + ".checker"))
         || "true".equals(System.getenv("ELASTIC_OTEL_JAVA_DISABLE_INSTRUMENTATIONS_CHECKER"))) {
-      new Thread(new OptionChecker(), "Elastic dynamic_instrumentation checker").start();
+      Thread checker = new Thread(new OptionChecker(), "Elastic dynamic_instrumentation checker");
+      checker.setDaemon(true);
+      checker.start();
     }
   }
 
   static class OptionChecker implements Runnable {
     private Map<String, Boolean> alreadyDisabled = new HashMap<>();
 
+    // Note that if the property and the API are both used to specify enablement
+    // for a particular instrument, and this thread is executing, the property
+    // will take priority if the instrument is in the property - by virtue of running
+    // more frequently; but won't if the instrument is removed from the property!
+    // TODO define priority of enablement by source of disabler
     @Override
     public void run() {
       while (true) {
--- a/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
+++ b/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
@@ -79,7 +79,7 @@ public class DynamicInstrumentation {
     }
   }
 
-  private static Object call(String methodname, Object target, Object arg1, Class<?> arg1Class) {
+  private static <T> Object call(String methodname, Object target, T arg1, Class<? super T> arg1Class) {
     try {
       Method method = target.getClass().getDeclaredMethod(methodname, arg1Class);
       method.setAccessible(true);
--- a/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
+++ b/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
@@ -79,7 +79,8 @@ public class DynamicInstrumentation {
     }
   }
 
-  private static <T> Object call(String methodname, Object target, T arg1, Class<? super T> arg1Class) {
+  private static <T> Object call(
+      String methodname, Object target, T arg1, Class<? super T> arg1Class) {
     try {
       Method method = target.getClass().getDeclaredMethod(methodname, arg1Class);
       method.setAccessible(true);
--- a/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
+++ b/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
@@ -31,8 +31,10 @@ import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -207,7 +209,7 @@ public class DynamicInstrumentation {
   }
 
   static class OptionChecker implements Runnable {
-    ConcurrentMap<String, Boolean> alreadyDisabled = new ConcurrentHashMap<>();
+    private Map<String, Boolean> alreadyDisabled = new HashMap<>();
 
     @Override
     public void run() {
--- a/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
+++ b/custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
@@ -1,0 +1,259 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.otel.config;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Tracer;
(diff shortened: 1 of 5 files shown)


custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
```diff
@@ -0,0 +1,261 @@
...
+      activatedTracers = null;
+    }
+    tracerSdkComponentRegistry
+        .getComponents()
+        .forEach(
+            sdkTracer -> {
+              try {
+                InstrumentationScopeInfo instrumentationScopeInfo =
+                    getInstrumentationScopeInfo(sdkTracer);
+                TracerConfig tConfig = getTracerConfig(finalProvider, instrumentationScopeInfo);
```
LikeTheSalad, line 149: Given the tools we have available upstream, I can see why we need to do the config update this way, though I was wondering (not sure if it has been discussed) if instead of evaluating a field [here](https://github.com/open-telemetry/opentelemetry-java/blob/eb53fe3a614a5feb6ddd74d93b43722a0033fffe/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracer.java#L38), we could instead change that line by something like `if (!tracerConfig.isEnabled()) {` instead? That way we might not need to keep a background thread alive constantly checking for config changes and applying them one by one by getting a new config and then resetting the tracerEnabled field as we could do the check reactively whenever `spanBuilder` is called.

I guess a downside might be that, depending on what's done in the `TracerConfig.isEnabled()` impl, it might cause some delays, though a good implementation shouldn't take long to return.

LikeTheSalad: If the delay concerns are a blocker, I think another alternative might be to make `TracerConfig` observable and then make the `SdkTracer` impl subscribe to changes in it, updating its `tracerEnabled` field whenever TracerConfig notifies about a change.

jackshirazi: > delay concerns are a blocker

They are not a blocker. It does not need to be immediate, it's fine for the instrumentation to be disabled but that not take effect for a while. Also, adding an observer-subscription model doesn't change the delay, because the delay is from `SdkTracer.tracerEnabled` being non-volatile. callbacks wouldn't change anything because the state change needs to propagate across threads and only making it volatile would speed that. I originally had it volatile in the PR that made it non-final, but the maintainers decided to go with a non-volatile implementation, at least for now

jackshirazi: > background thread alive constantly checking for config changes

The background thread checker is no on by default. I expect normal operation for something else - eg an OpAMP client - to do regular checking from central config (and maybe property and config file, to be decided) and update the instrumentation enablement from that. The background thread checker is there mainly so that testing can be done cleanly, though it is fully operational if someone wanted to use it in production

LikeTheSalad: > They are not a blocker. It does not need to be immediate

Sorry I think I didn't properly explain what I meant by delay, I was more talking about the possibility that by changing [this condition](https://github.com/open-telemetry/opentelemetry-java/blob/eb53fe3a614a5feb6ddd74d93b43722a0033fffe/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/SdkTracer.java#L38) to always query the config by calling `TracerConfig.isEnabled()` instead of evaluating a field, then it could cause a performance overhead to the instrumented app in case that the provided `TracerConfig` impl was poorly created, for example, let's say a `TracerConfig` impl reads a file every time its `isEnabled` method is called, then that could block the span creation process for a little while, causing the host app performance to get affected every time it creates a span. On the other hand, if we decide to keep the field then I agree we should make it volatile, though I was more of the opinion of removing that field and instead evaluating the config on every span creation.

LikeTheSalad: Though it makes sense what you mentioned about the OpAMP client triggering the update process when it gets a new config. I guess with the alternatives I mentioned I was also trying to make the "update process" something that couldn't be triggered from the outside since I'm not sure if it'd be desirable for any part of the code that might have access to the Otel instance (including via the GlobalOpenTelemetry singleton) to make config changes. However, if it's only a matter of adding the `updateTracerConfigurations()` method without params, but only as a sort of "notify changes" signal to get the tracer to re-evaluate its existing config provider, then it should be fine.

jackshirazi: > in case that the provided `TracerConfig` impl was poorly created,

I tried to open the TracerConfig to be implementable by the user, but the maintainers preferred to keep it as is, essentially a boolean. So atm it's not possible for it to be inefficient

custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
```diff
@@ -0,0 +1,261 @@
...
+      Method method = target.getClass().getDeclaredMethod(methodname);
+      method.setAccessible(true);
+      return method.invoke(target);
+    } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+      throw new IllegalStateException(
+          "Error calling " + methodname + " on " + target.getClass(), e);
+    }
+  }
+
+  private static Object call(String methodname, Object target, Object arg1, Class<?> arg1Class) {
```
JonasKunz, line 82:
```suggestion
  private static <T> Object call(String methodname, Object target, T arg1, Class<? super T> arg1Class) {
```

[Nit] introduces a minimal bit of type safety

custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
```diff
@@ -0,0 +1,261 @@
...
+
+    public void put(InstrumentationScopeInfo scope, TracerConfig tracerConfig) {
+      map.put(scope.getName(), tracerConfig);
+    }
+  }
+
+  static {
+    if ("true".equals(System.getProperty(INSTRUMENTATION_DISABLE_OPTION + ".checker"))
+        || "true".equals(System.getenv("ELASTIC_OTEL_JAVA_DISABLE_INSTRUMENTATIONS_CHECKER"))) {
+      new Thread(new OptionChecker(), "Elastic dynamic_instrumentation checker").start();
```
JonasKunz, line 207: This will start a non daemon thread, which is a bad Idea IMO.

I'd suggest to do the following:
* Move the checker start logic into `DynamicInstrumentation.setTracerConfigurator`
* Add the `ConfigProperties` as argument to that method: This means you don't have to check env variables / proeprties and to the conversion yourself
* There, create a single-threaded daemon executor service with your periodic checker
* For a proper shutdown, I've used a trick before: On the `TracerProviderBuilder`, register a SpanProcessor which does nothing. It will still receive a shutdown event though, which can be used to properly shutdown the executor

custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+  }
+
+  static class OptionChecker implements Runnable {
+    private Map<String, Boolean> alreadyDisabled = new HashMap<>();
```
JonasKunz, line 212: This algorithm is stateful due to carrying over `alreadyDisabled` over the loop iterations.

I think this can be greatly simplified because we have that state already in another place, namely `UpdatableConfigurator.map`:

What I would do is something like adding a `boolean setOnlyDisabled(Set<String> scopeNames)` method to `UpdateableConfigurator` which:
 * Reenables all `TraceConfigs` in `map` which are not mentioned in `scopeNames`
 * Makes all elements in `scopeNames` disabled in `map`
 
 As a result, you can now just pass the the desired target state (the list of disabled instrumentations coming form the config) to it periodically and dependent on the return value update the tracers.

jackshirazi: That would make it beautifully simple! However that assumes only this checker (ie only this property) sets disablement, which is not the case. Example scenario: central config disables instrument X. `disable_instrumentations` property remains empty. This thread executes - result would be that central config client would disable X and then this thread would re-enable it almost immediately.

jackshirazi: But that does identify that there is a priority conflict in which decides enablement if both are used for the same instrumentation. As this thread is really only here for testing the functionality (although it can be used in production), I'll leave it as is since deciding priority will make it more complicated. I'll add a note about that though

custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
jackshirazi: > - Move the checker start logic into DynamicInstrumentation.setTracerConfigurator

Although that works now, it just shifts needing to spawn it in a static initializer to a later implementation because setTracerConfigurator is really just a proxy for `SdkTracerProviderBuilder.setTracerConfigurator` until that latter becomes public, at which point `DynamicInstrumentation.setTracerConfigurator` disappears entirely - as do most of the DynamicInstrumentation methods.

I think it's fine to just make it a daemon thread - good catch! done that now

custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
JonasKunz: > Example scenario: central config disables instrument X. disable_instrumentations property remains empty

I was expecting that the config merge happens beforehand and that this thread operates on the already merged conflicts and therefore is the only one performing the updates. However, as this is not the case you are totally right.

custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
JonasKunz: Good for me, we can have another look later when things get more stable/public

JonasKunz (approved):

LikeTheSalad (approved):

PR #428, opened 2024-10-09
Test with JDK 23

JonasKunz:


jackshirazi:
lol, I thought you meant we needed to add the full framework rather than just this bump!

jackshirazi (approved):

JonasKunz:
Finally found the root cause for the flaky `testVirtualThreadsExcludedByDefault`:
That test would sometimes fail because the native profiling correlation TLS would be non-null even though it should be.

The reason for this was an improper cleanup in `testVirtualThreadsStoragePropagatedWithMounts`:
This test would span lots of virtual thread and populate the carrier thread TLS. After the test the syncinc of virtual thread TLS with the carrier thread TLS would be disabled. The problem now was that we disabled the sync without actually waiting for all virtual threads to be terminated and therefore unmounted. As a result, for some carrier the TLS would not be cleaned up properly, leaking into the `testVirtualThreadsExcludedByDefault` test.

jackshirazi (approved):

PR #454, opened 2024-11-19
Add changelog generation to release process

JonasKunz:
* Adds a new `CHANGELOG.md` to the repository
 * The contents of this file will be used as release notes on release and cleared afterwards
 * In addition, the versions of relevant dependencies are appended to the release notes automatically
 
 Example:
 
image

I've tested the release note generation steps using GH actions in my fork of the repo.


.github/workflows/release-step-3.yml
```diff
@@ -200,11 +200,35 @@ jobs:
...
+        with:
+          command: ""
+      - name: Print Release Notes
+        id: print_release_notes
+        run: |
+          echo 'notes<<RELNOTESEOF' >> $GITHUB_OUTPUT
+          cat CHANGELOG.next-release.md >> $GITHUB_OUTPUT
+          printf '\nThis release is based on the following upstream versions:\n\n' >> $GITHUB_OUTPUT
+          ./gradlew -q printUpstreamDependenciesMarkdown >> $GITHUB_OUTPUT
+          echo 'RELNOTESEOF' >> $GITHUB_OUTPUT
```
SylvainJuge, line 224: [minor] could it be simpler to implement this completely as a gradle task ?

SylvainJuge (approved):

.github/workflows/release-step-3.yml
JonasKunz: I don't think it would be really simpler because then the gradle task would have to do the file handling, which is more code than `cat CHANGELOG.next-release.md`.
The multiline support code `echo 'notes<<RELNOTESEOF' >> $GITHUB_OUTPUT` and `echo 'RELNOTESEOF' >> $GITHUB_OUTPUT ` would still be required, so we wouldn't save much

PR #474, opened 2024-12-06
add stop-sending option

jackshirazi:


custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
```diff
@@ -0,0 +1,77 @@
...
+
+  private volatile boolean sendingLogs = true;
+  private final LogRecordExporter delegate;
+
+  public static ElasticLogRecordExporter getInstance() {
+    return INSTANCE.get();
+  }
+
+  static ElasticLogRecordExporter createCustomInstance(LogRecordExporter exporter) {
+    INSTANCE.set(new ElasticLogRecordExporter(exporter));
```
jackshirazi, line 38: For reviewer. This would ideally be a single instance that doesn't get recreated, but the test framework fails if you do that, so it needs to be recreated each time this method is called

custom/src/main/java/co/elastic/otel/ElasticMetricExporter.java
```diff
@@ -0,0 +1,90 @@
...
+    sendingMetrics = send;
+  }
+
+  public boolean sendingMetrics() {
+    return sendingMetrics;
+  }
+
+  @Override
+  public AggregationTemporality getAggregationTemporality(InstrumentType instrumentType) {
+    return AggregationTemporality.DELTA;
```
jackshirazi, line 59: Retained from when we previously had this class, not sure if we do want this

custom/src/main/java/co/elastic/otel/config/DynamicConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+ * under the License.
+ */
+package co.elastic.otel.config;
+
+import co.elastic.otel.ElasticLogRecordExporter;
+import co.elastic.otel.ElasticMetricExporter;
+import co.elastic.otel.ElasticSpanExporter;
+
+public class DynamicConfiguration {
+  private static DynamicConfiguration INSTANCE = new DynamicConfiguration();
```
jackshirazi, line 26: For reviewer, this class is essentially what the OpAMP client would use, or extend or be implemented into

custom/src/main/java/co/elastic/otel/config/DynamicConfigurationPropertyChecker.java
```diff
@@ -0,0 +1,133 @@
...
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class DynamicConfigurationPropertyChecker implements Runnable {
+  private static Thread CheckerThread;
```
jackshirazi, line 29: This functionality was in the DynamicInstrumentation class, but it's clear that we need to use this for testing for any dynamic config we implement, so I've refactored it into it's own class

custom/src/main/java/co/elastic/otel/config/DynamicInstrumentation.java
```diff
@@ -130,7 +118,7 @@ private static InstrumentationScopeInfo getInstrumentationScopeInfo(Tracer sdkTr
   //                    getTracerConfig(sdkTracer.getInstrumentationScopeInfo())));
   // where SdkTracer.updateTracerConfig(TracerConfig tracerConfig) is equivalent to
   //  this.tracerEnabled = tracerConfig.isEnabled();
-  private static void updateTracerConfigurations(TracerProvider provider) {
+  static void updateTracerConfigurations(TracerProvider provider) {
```
jackshirazi, line 121: All public methods have been refactored to the DynamicConfiguration class, this class only holds methods that should ultimately disappear when the SDK APIs become non-experimental

custom/src/main/java/co/elastic/otel/config/DynamicConfigurationPropertyChecker.java
```diff
@@ -0,0 +1,133 @@
...
+                System.getProperty(
+                    DynamicConfiguration.INSTRUMENTATION_DISABLE_OPTION + ".checker"))
+        || "true"
+            .equals(
+                System.getenv("ELASTIC_OTEL_JAVA_EXPERIMENTAL_DISABLE_INSTRUMENTATIONS_CHECKER"))) {
+      try {
+        Interval =
+            Long.parseLong(
+                System.getenv(
+                    "ELASTIC_OTEL_JAVA_EXPERIMENTAL_DISABLE_INSTRUMENTATIONS_CHECKER_INTERVAL_MS"));
```
jackshirazi, line 49: added to speed up testing

testing-common/src/main/java/co/elastic/otel/testing/OtelReflectionUtils.java
```diff
@@ -54,8 +54,11 @@ public static void shutdownAndResetGlobalOtel() {
         (OpenTelemetry) readField(GlobalOpenTelemetry.class, null, "globalOpenTelemetry");
     if (otel != null) {
       // unwrap from obfuscated opentelemetry
-      OpenTelemetrySdk sdk = (OpenTelemetrySdk) readField(otel, "delegate");
-      sdk.close();
+      Object sdkObject = readField(otel, "delegate");
```
jackshirazi, line 57: This was failing after custom exporters are added, the instance is no longer an OpenTelemetrySdk instance

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
```diff
@@ -0,0 +1,77 @@
...
+  public boolean sendingLogs() {
+    return sendingLogs;
+  }
+
+  @Override
+  public CompletableResultCode export(Collection<LogRecordData> collection) {
+    if (sendingLogs) {
+      return delegate.export(collection);
+    } else {
+      return CompletableResultCode.ofFailure();
```
LikeTheSalad, line 59: Maybe I'm overthinking this, though if the export wasn't carried out due to our config flag, isn't that "as expected"? meaning that the operation was successful in a way? I'm asking because I can see a case where some exporter (the disk exporter is one for example) might decide to try again later if it receives an "export failed" result, and I'm not sure if we'd like that to happen for this case. I'm fine with leaving it like this btw, but it's something I'd like to mention in case it wasn't considered before.

custom/src/main/java/co/elastic/otel/config/DynamicConfigurationPropertyChecker.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class DynamicConfigurationPropertyChecker implements Runnable {
+  private static Thread CheckerThread;
+  private static long Interval = 1000;
```
LikeTheSalad, line 30: nit: shouldn't we use lower camel case?

testing-common/src/main/java/co/elastic/otel/testing/OtelReflectionUtils.java
JonasKunz: Is this still the case in the current state?
I removed the `if` check and the tests still run fine for me.

LikeTheSalad (approved):

custom/src/main/java/co/elastic/otel/config/DynamicConfigurationPropertyChecker.java
JonasKunz: Interesting, would have expected this to be complained about by spotless

custom/src/main/java/co/elastic/otel/ElasticAutoConfigurationCustomizerProvider.java
```diff
@@ -48,12 +49,21 @@ public class ElasticAutoConfigurationCustomizerProvider
 
   @Override
   public void customize(AutoConfigurationCustomizer autoConfiguration) {
+    autoConfiguration.addMetricExporterCustomizer(
+        (metricexporter, configProperties) ->
+            ElasticMetricExporter.createCustomInstance(metricexporter));
+    autoConfiguration.addSpanExporterCustomizer(
+        (spanExporter, configProperties) -> ElasticSpanExporter.createCustomInstance(spanExporter));
```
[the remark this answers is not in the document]
jackshirazi: My understanding is that this is a request for an "emergency stop" option in cases where there is something overwhelming the app/system. So yes, I'd say we'd want to stop all output. later on we might get requests for logging to continue, but we should treat that as a separate enhancement

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+package co.elastic.otel;
+
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.logs.data.LogRecordData;
+import io.opentelemetry.sdk.logs.export.LogRecordExporter;
+import java.util.Collection;
+import java.util.concurrent.atomic.AtomicReference;
+
+public class ElasticLogRecordExporter implements LogRecordExporter {
```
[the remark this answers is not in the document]
jackshirazi: Elastic is used here as the prepend to distinguish it from vanilla. I'm fine with changing the name but Disableable is for this very specific functionality and I expect we'll have other uses. Maybe `EDOT*Exporter`?

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
```diff
@@ -0,0 +1,77 @@
...
+  private static final AtomicReference<ElasticLogRecordExporter> INSTANCE = new AtomicReference<>();
```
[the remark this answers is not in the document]
jackshirazi: You are right, originally I had this as a singleton so used the atomic getandset to ensure it was only set once. But now it only needs visibility

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
jackshirazi: I really don't know, I was 50/50 on it myself. I did a https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-java%20CompletableResultCode.ofFailure&type=code check and the SDK sends failure on shutdown, stopping is a reversible shutdown so I went with failure. I'm still 50/50, I'll ask on the otel Java channel

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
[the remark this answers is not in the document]
JonasKunz: >  I expect we'll have other uses
I would expect those other uses then to be in separate exporter wrappers to have a clear separation of concerns. The fact that it is not vanilla is shown via the `co.elastic.otel` package this class lives in.

custom/src/main/java/co/elastic/otel/config/DynamicConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+  public static final String DISABLE_SEND_OPTION = "elastic.otel.java.experimental.disable_send";
+  public static final String INSTRUMENTATION_DISABLE_OPTION =
+      "elastic.otel.java.experimental.disable_instrumentations";
+
+  private Boolean recoverySendSpansState;
+  private Boolean recoverySendLogsState;
+  private Boolean recoverySendMetricsState;
+
+  private boolean initSendingStates() {
+    if (recoverySendSpansState == null && ElasticSpanExporter.getInstance() != null) {
```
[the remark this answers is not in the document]
jackshirazi: yes, good call

custom/src/main/java/co/elastic/otel/config/DynamicConfigurationPropertyChecker.java
jackshirazi: I've always used uppercase first letter on static fields, but I'm find with lower-casing it

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
jackshirazi: Reply from Lauri
```
BatchSpanProcessor logs a debug message on failed export, on successful export it increments a metric. Based on that I'd say that your exporter should return success.
```
so given everyone seems to feel it's success, I'll flip to that

LikeTheSalad: It wasn't fully clear to me either so I'm glad we could get someone like Lauri to confirm! Thank you.

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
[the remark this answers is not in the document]
jackshirazi: Good points, I agree now. `Dynamic*Exporter`? `Blockable*Exporter`? Disableable and Stoppable sound strange (though ironically unstoppable is perfectly good and quite commonly used)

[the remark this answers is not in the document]
JonasKunz: `Blockable` sounds good

smoke-tests/src/test/java/com/example/javaagent/smoketest/DynamicConfigSmokeTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @AfterAll
+  public static void end() {
+    stopApp();
+  }
+
+  @AfterEach
+  public void endTest() throws InterruptedException {
+    doRequest(getUrl("/dynamicconfig/reset"), okResponseBody("reset"));
+    Thread.sleep(500L); // give the reset time to be applied
```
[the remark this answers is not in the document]
jackshirazi: there was a reason which is that I prefer the smoke tests and find the integration tests non-transparent in how they work. Though I take your point they are easier to maintain. And these tests **are** actually integration testst. I've done an initial pass and found a bug the integration test identified that the smoke tests didn't, so I'll map them over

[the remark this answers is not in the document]
jackshirazi: per discussion (the integration test is unable to handle the exporter wrapping correctly because it has a custom exporter that doesn't go through configuration), staying with the smoke test

testing-common/src/main/java/co/elastic/otel/testing/OtelReflectionUtils.java
jackshirazi: reverted in c7b825d

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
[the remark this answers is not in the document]
jackshirazi: fixed in c7b825d

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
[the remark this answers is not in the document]
jackshirazi: applied in c7b825d

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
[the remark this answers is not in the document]
jackshirazi: applied in c7b825d

custom/src/main/java/co/elastic/otel/ElasticLogRecordExporter.java
jackshirazi: applied in c7b825d

JonasKunz (approved):

PR #497, opened 2025-01-15
Add experimental instrumentation for OpenAI client

JonasKunz:
Adds an experimental instrumentation module for the [official OpenAI Java client](https://github.com/openai/openai-java).


instrumentation/openai-client-instrumentation/src/main/java/co/elastic/otel/openai/wrappers/InstrumentationSettings.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class InstrumentationSettings {
+
+  final boolean emitEvents;
+  final boolean captureMessageContent;
+
+  // we do not directly have access to the client baseUrl after construction, therefore we need to
+  // remember it
+  final String serverAddress;
+  final Long serverPort;
```
SylvainJuge, line 34: [minor] is there any reason not to make those private ?

instrumentation/openai-client-instrumentation/src/main/java/co/elastic/otel/openai/wrappers/InstrumentedChatCompletionService.java
```diff
@@ -0,0 +1,271 @@
...
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.jetbrains.annotations.NotNull;
```
SylvainJuge, line 57: same as previously, should we `@Nullable` instead ?

instrumentation/openai-client-instrumentation/src/test/java/co/elastic/otel/openai/EmbeddingsTest.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+package co.elastic.otel.openai;
+
+import static io.opentelemetry.sdk.testing.assertj.OpenTelemetryAssertions.assertThat;
+import static io.opentelemetry.sdk.testing.assertj.OpenTelemetryAssertions.equalTo;
+import static io.opentelemetry.semconv.ServerAttributes.SERVER_ADDRESS;
+import static io.opentelemetry.semconv.ServerAttributes.SERVER_PORT;
+import static io.opentelemetry.semconv.incubating.GenAiIncubatingAttributes.GEN_AI_OPERATION_NAME;
+import static io.opentelemetry.semconv.incubating.GenAiIncubatingAttributes.GEN_AI_REQUEST_ENCODING_FORMATS;
+import static io.opentelemetry.semconv.incubating.GenAiIncubatingAttributes.GEN_AI_REQUEST_MODEL;
```
SylvainJuge, line 27: This indirectly verifies that the incubating semconv and their equivalents stay aligned, so we cold skip having them tested explicitly if all their equivalents are being used in tests. This is how this is being verified in the upstream instrumentation and contrib, there is no explicit test for the semconv attributes.

instrumentation/openai-client-instrumentation/src/test/java/co/elastic/otel/openai/ValAssert.java
```diff
@@ -0,0 +1,191 @@
...
+import io.opentelemetry.api.common.ValueType;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+
+public interface ValAssert extends Consumer<Value<?>>, BiConsumer<String, Value<?>> {
```
SylvainJuge, line 34: I know this might require a bit of effort, but it might be better to do a proper assertj extension as we have in our classic agent, the downside however is having to rewrite assertions using a different API. Given it's about otel types, those could even be contributed upstream and reused.

instrumentation/openai-client-instrumentation/src/test/java/co/elastic/otel/openai/ChatTest.java
```diff
@@ -0,0 +1,2387 @@
...
+    assertThat(testing.logRecords())
+        .hasSize(2)
+        .anySatisfy(
+            log -> {
+              assertThat(log)
+                  .hasAttributesSatisfying(
+                      attr ->
+                          assertThat(attr)
+                              .containsEntry(GEN_AI_SYSTEM, "openai")
+                              .containsEntry("event.name", "gen_ai.user.message"))
```
SylvainJuge, line 177: [question] does values of `event.name` are part of the semantic conventions ?

SylvainJuge (approved):

buildSrc/src/main/kotlin/elastic-otel.instrumentation-conventions.gradle.kts
```diff
@@ -0,0 +1,91 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
```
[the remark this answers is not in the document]
JonasKunz: They kept the original package names, I think for backwards-compatibility of gradle scripts

instrumentation/openai-client-instrumentation/src/main/java/co/elastic/otel/openai/wrappers/EventLoggingStreamedResponse.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+package co.elastic.otel.openai.wrappers;
+
+import com.openai.core.http.StreamResponse;
+import com.openai.models.ChatCompletionChunk;
+import io.opentelemetry.context.Context;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Stream;
+import org.jetbrains.annotations.NotNull;
```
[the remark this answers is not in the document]
JonasKunz: These made it into here because the OpenAI client seems to use them, I'll just remove them

instrumentation/openai-client-instrumentation/src/main/java/co/elastic/otel/openai/wrappers/GenAiAttributes.java
```diff
@@ -0,0 +1,73 @@
...
+ * under the License.
+ */
+package co.elastic.otel.openai.wrappers;
+
+import io.opentelemetry.api.common.AttributeKey;
+import java.util.List;
+
+public class GenAiAttributes {
+
+  // TODO: ensure with a test that the attributes don't go out of sync with semconv
```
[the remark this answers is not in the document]
JonasKunz: Agree with your other message that this is actually already covered by the tests, so just removing this TODO

instrumentation/openai-client-instrumentation/src/main/java/co/elastic/otel/openai/wrappers/InstrumentationSettings.java
JonasKunz: This is required for testing, I'll add a comment

instrumentation/openai-client-instrumentation/src/main/java/co/elastic/otel/openai/wrappers/InstrumentedChatCompletionService.java
JonasKunz: Removed

instrumentation/openai-client-instrumentation/src/test/java/co/elastic/otel/openai/ChatTest.java
JonasKunz: They are part of the semantic conventions, but not yet of the auto-generated libs it seems.

instrumentation/openai-client-instrumentation/src/test/java/co/elastic/otel/openai/ValAssert.java
JonasKunz: I agree that this would be the ideal solution, but also a lot of effort to get readable assertions. Are you okay with keeping it as is for now?

instrumentation/openai-client-instrumentation/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+  testImplementation(catalog.openaiClient)
+
+  testImplementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.18.2")
+  testImplementation("com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.18.2")
+  testImplementation("org.slf4j:slf4j-simple:2.0.16")
+  testImplementation(catalog.wiremock)
+}
+
+muzzle {
+  // TODO: setup muzzle to check older versions of openAI client
```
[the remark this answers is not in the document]
JonasKunz: issue: https://github.com/elastic/elastic-otel-java/issues/499

instrumentation/openai-client-instrumentation/src/test/java/co/elastic/otel/openai/ValAssert.java
SylvainJuge: Yes, that's fine. Contributing this upstream would probably be a better choice anyway, then reusing them here would be likely easier.

SylvainJuge (approved):

PR #531, opened 2025-02-05
Fix instrumentation support for OpenAI client 0.14+

JonasKunz:
The OpenAI java client 0.14 had several breaking changes due to some methods we used being renamed.
The other logic is still all the same, that's why I decided to simply encapsulate those methods in an `ApiAdapter` and provide implementations for each the pre and post 0.14.0 versions.

For on how to install the instrumentation I considered two alternatives:
 * Stay with a single instrumentation module and simply load the correct adapter at runtime after inspecting via reflection which OpenAI client version is present
 * Use two instrumentations modules, each loading a specific adapter
 
 While the second option introduces slightly more code, it has the benefit of muzzle working correctly: For each module, muzzle will pick up the OpenAI client methods used in the `ApiAdapterImpl`s and will therefore ensure that future breaking changes to those are correctly detected. That's why I went for two instrumentation modules here.


custom/build.gradle.kts
```diff
@@ -3,7 +3,8 @@ plugins {
 }
 
 val instrumentations = listOf<String>(
-  ":instrumentation:openai-client-instrumentation"
+  ":instrumentation:openai-client-instrumentation:instrumentation-0.13.0",
+  ":instrumentation:openai-client-instrumentation:instrumentation-latest"
```
SylvainJuge, line 7: [question] where does the `latest` comes from, is it a feature of the gradle dependencies to allow easily getting the latest ?

instrumentation/openai-client-instrumentation/common/src/main/java/co/elastic/otel/openai/wrappers/InstrumentedChatCompletionService.java
```diff
@@ -137,11 +137,9 @@ public void onStart(
...
                       .ifPresent(
                           val -> {
-                            if (val.isResponseFormatText()) {
-                              attributes.put(
-                                  GEN_AI_OPENAI_REQUEST_RESPONSE_FORMAT,
-                                  val.asResponseFormatText()._type().toString());
-                            }
+                            attributes.put(
+                                GEN_AI_OPENAI_REQUEST_RESPONSE_FORMAT,
+                                ApiAdapter.get().extractType(val));
```
SylvainJuge, line 142: If I remember correctly adding an attribute with a `null` value is ignored, so if the response is not text then we should not add anything, is that correct ?

instrumentation/openai-client-instrumentation/instrumentation-0.13.0/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
+plugins {
+  alias(catalog.plugins.muzzleGeneration)
+  alias(catalog.plugins.muzzleCheck)
+  id("elastic-otel.instrumentation-conventions")
+}
+
+val openAiVersion = "0.13.0"; // DO NOT UPGRADE
```
SylvainJuge, line 7: Is it something that renovate/dependabot would attempt to upgrade ? If so I think there are ways like splitting the string in parts to prevent this.

instrumentation/openai-client-instrumentation/instrumentation-0.13.0/src/main/java/co/elastic/otel/openai/v0_13_0/ApiAdapterImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+    return null;
+  }
+
+  @Override
+  public String asText(ChatCompletionUserMessageParam.Content content) {
+    if (content.isTextContent()) {
+      return content.asTextContent();
+    }
+    return "";
```
SylvainJuge, line 91: [minor] maybe use a static method to avoid duplication.

instrumentation/openai-client-instrumentation/instrumentation-latest/src/main/java/co/elastic/otel/openai/latest/ApiAdapterImpl.java
```diff
@@ -0,0 +1,104 @@
...
+    if (val.isText()) {
+      return val.asText()._type().toString();
+    }
+    if (val.isJsonObject()) {
+      return val.asJsonObject()._type().toString();
+    }
+    if (val.isJsonSchema()) {
+      return val.asJsonSchema()._type().toString();
+    }
+    return "";
```
SylvainJuge, line 102: is it expected that we return an empty string here, other `extractXxxx` methods return `null` so this seems a bit inconsistent.

settings.gradle.kts
```diff
@@ -17,7 +17,10 @@ include("agentextension")
 include("bootstrap")
 include("custom")
 include("instrumentation")
-include("instrumentation:openai-client-instrumentation")
+include("instrumentation:openai-client-instrumentation:common")
+include("instrumentation:openai-client-instrumentation:testing-common")
+include("instrumentation:openai-client-instrumentation:instrumentation-0.13.0")
+include("instrumentation:openai-client-instrumentation:instrumentation-latest")
```
SylvainJuge, line 23: [question] In the upstream instrumentation, I think the recommended pattern is to always include the minimum version supported for a module, both in the gradle module name and in the code packages. There is this draft PR upstream: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/12965 so maybe starting to use this for our own instrumentations would be a good start as it's less painful than changing it later.

instrumentation/openai-client-instrumentation/instrumentation-0.13.0/build.gradle.kts
JonasKunz: Good idea. I don't know if renovate is smart enough for this. We can just keep it like this for now and see if renovate detects it and attempts an upgrade

instrumentation/openai-client-instrumentation/common/src/main/java/co/elastic/otel/openai/wrappers/InstrumentedChatCompletionService.java
JonasKunz: I checked the implementation, you are correct. Nonetheless we should introduce a null check I think for clarity and make `extractType` return `null` if it can't extract a type.

> so if the response is not text then we should not add anything, is that correct
It should return a string identifying the type whenever it is available. The previous implementation did this only for text, so this is a slight enhancement I smuggled into this PR.

JonasKunz: Done in [106a90c](https://github.com/elastic/elastic-otel-java/pull/531/commits/106a90c8e32b287f761aa168da49ba6077a73823).

instrumentation/openai-client-instrumentation/instrumentation-latest/src/main/java/co/elastic/otel/openai/latest/ApiAdapterImpl.java
JonasKunz: It should be `null`, fixed in [106a90c](https://github.com/elastic/elastic-otel-java/pull/531/commits/106a90c8e32b287f761aa168da49ba6077a73823).

instrumentation/openai-client-instrumentation/instrumentation-0.13.0/src/main/java/co/elastic/otel/openai/v0_13_0/OpenAiOkHttpClientBuilderInstrumentation.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public static class AdviceClass {
+
+    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+    @Advice.AssignReturned.ToReturned
+    public static OpenAIClient onExit(
+        @Advice.Return OpenAIClient result, @Advice.FieldValue("baseUrl") String baseUrl) {
+      // This initialization has two purposes:
+      // Initialize the correct adapter AND ensure that it is picked up by muzzle
```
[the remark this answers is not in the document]
JonasKunz: > We could have moved this to a static init in InstrumentedOpenAiClient, however since this one is in the common part we can't.

Yeah this exactly the reason for the duplication into two separate instrumentation modules: They use different `ApiAdapterImpl`s (distinguished by different package names).

instrumentation/openai-client-instrumentation/instrumentation-0.13.0/src/main/java/co/elastic/otel/openai/v0_13_0/ApiAdapterImpl.java
JonasKunz: I switched to a ternary operator to make thinks more concise now. Other than that, I don't think we can extract anything here, all the `Content` types across the methods are differen.

SylvainJuge (approved):

PR #536, opened 2025-02-06
Prevent incorrect renovate upgrades: second attempt

JonasKunz:


SylvainJuge (approved):
LGTM to start simple, we could make it fancy by adding a gradle task that checks all the wildcard-less entries in renovate.json are matching actual files, but that's probably too much to only save us of a few PRs if we ever rename it.

PR #537, opened 2025-02-06
chore(renovate): don't automagically upgrade OpenAI SDK

jamietanna:
- **Revert "Prevent renovate from upgrading intentionally fixed openai version (#534)"**
- **chore(renovate): don't automagically upgrade OpenAI SDK**

### DIFF
--- a/instrumentation/openai-client-instrumentation/instrumentation-0.2/build.gradle.kts
+++ b/instrumentation/openai-client-instrumentation/instrumentation-0.2/build.gradle.kts
@@ -5,7 +5,7 @@ plugins {
   id("elastic-otel.instrumentation-conventions")
 }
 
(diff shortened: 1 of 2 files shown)


jamietanna:
Would also need to pull in https://github.com/elastic/elastic-otel-java/pull/536

JonasKunz:
Unfortunately the suggested approach won't work for us. We want automatic upgrades of the OpenAI client, just not on this particular subproject.

That project is intended to test with the fixed `0.13.0` version. We have another sub-project which tests against the latest version and should be upgraded.

I don't think there is any better way of doing this other than excluding that specific subproject, is there?

JonasKunz:
See for example this bump:
https://github.com/elastic/elastic-otel-java/pull/530

That's the one we still need.

JonasKunz:
Closing due to the remarks in my previous comments

PR #539, opened 2025-02-11
Fix support for OpenAI developer messages

### LINKED ISSUE #532: Add support for OpenAI "Developer Messages" introduced in openAI java client 0.8
While implementing #531, I discoveres that openAI added [developer messages](https://lunary.ai/blog/openai-developer-role#what-are-developer-messages) as replacement for system messages.
We'll need to decide on how to handle them, because currently the semantic conventions don't cover those.
At the moment the instrumentation will fail if event logging is enabled, because it encounteres an unknown type.
Some possible alternatives:
 * Don't report as an event and wait for semconv
 * Report as [system.message](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-events/#event-gen_aisystemmessage) for now

JonasKunz:
Closes #532 . For now, these messages will simply be reported as `gen_ai.system.message` events with `role` set to `developer`, which seems to be kind of sem conv compliant. We'll have to revisit later anyway for future semconv changes (e.g. https://github.com/open-telemetry/semantic-conventions/issues/1851), but this at least keeps things working for now.


instrumentation/openai-client-instrumentation/common/src/main/java/co/elastic/otel/openai/wrappers/ChatCompletionEventsHelper.java
```diff
@@ -102,7 +115,8 @@ public static void emitPromptLogEvents(
           bodyBuilder.put("id", toolMsg.toolCallId());
         }
       } else {
-        throw new IllegalStateException("Unhandled type : " + msg.getClass().getName());
+        LOG.log(Level.WARNING, "Unhandled OpenAI message type will be dropped: {0}", msg);
```
JonasKunz, line 118: This would previously cause the instrumentation to fail, which is a bit too strict.

instrumentation/openai-client-instrumentation/common/src/main/java/co/elastic/otel/openai/wrappers/ChatCompletionEventsHelper.java
```diff
@@ -228,7 +255,7 @@ private static LogRecordBuilder newEvent(String name) {
   private static Value<?> buildToolCallEventObject(ChatCompletionMessageToolCall call) {
     Map<String, Value<?>> result = new HashMap<>();
     result.put("id", Value.of(call.id()));
-    result.put("type", Value.of(call._type().toString()));
```
JonasKunz, line 231: This was not detected by muzzle: In some versions inbetween `_type()` was not present in the openAI lib.

instrumentation/openai-client-instrumentation/instrumentation-0.8/src/main/java/co/elastic/otel/openai/v0_2/ApiAdapterImpl.java
```diff
@@ -61,11 +65,11 @@ public String extractText(ChatCompletionContentPart part) {
...
   public String extractType(ChatCompletionCreateParams.ResponseFormat val) {
     if (val.isResponseFormatText()) {
-      return val.asResponseFormatText()._type().toString();
+      return "text";
     } else if (val.isResponseFormatJsonObject()) {
-      return val.asResponseFormatJsonObject()._type().toString();
+      return "json_object";
     } else if (val.isResponseFormatJsonSchema()) {
-      return val.asResponseFormatJsonSchema()._type().toString();
+      return "json_schema";
```
JonasKunz, line 72: For some versions `_type()` returns null here, so I instead hardcoded the strings for these older versions.

codefromthecrypt (approved):

PR #560, opened 2025-03-03
document micrometer metrics

SylvainJuge:
- document that micrometer metrics are disabled by default
- add configuration option to enable them
- add link to the other related options for micrometer metrics to otel conversion (in the upstream instrumentation).


jackshirazi (approved):

codefromthecrypt (approved):
Thanks for documenting this!

PR #561, opened 2025-03-04
fix upstream otel update

SylvainJuge:
Upstream agent update brings breaking changes from semconv which needs to be fixed by this PR.

- fix `host.id` and `service.namespace`
- fix removal of `GlobalEventLoggerProvider`
- remove `GlobalEventLoggerProvider` again
- remove more usages of `ResourceAttributes`

- contains commits from https://github.com/elastic/elastic-otel-java/pull/553 that was closed by automation.


gradle/libs.versions.toml
```diff
@@ -10,17 +10,17 @@ opentelemetryProto = "1.3.2-alpha"
...
 # updated from upstream agent with .ci/update-upstream.sh
-opentelemetryContribAlpha = "1.42.0-alpha"
+opentelemetryContribAlpha = "1.44.0-alpha"
 
 # otel semconv
 # updated from upstream agent with .ci/update-upstream.sh
 # While the semconv stable artifact is provided transitively by the agent, we still have to explicitly
 # reference the "incubating" version explicitly
-opentelemetrySemconvAlpha = "1.29.0-alpha"
+opentelemetrySemconvAlpha = "1.30.0-rc.1"
```
JonasKunz, line 23: This will also require a change to our gradle task printing the release notes, which currently has the `-alpha` suffix hardcoded. We should probably replace that with a `-.*` regex.
https://github.com/elastic/elastic-otel-java/blob/main/build.gradle.kts#L57

SylvainJuge: done in e950781fdd93d5d52a7d0e9192fa521628bdb914

JonasKunz (approved):
Thanks for taking care of this tedious work!

PR #564, opened 2025-03-05
Add support for OpenAI client 0.22 to 0.31

JonasKunz:
These versions
 * added some new methods in the chat-completion and embeddings API which we don't trace for now, but therefore require dynamic proxying
 * removed the `Exception` from the `close()` signature, which fortunately for us is a bytecode-backwards compatible change
 
 I don't think we need to check and add those new methods to our instrumentation now, we should revisit when the OpenAI client is out of beta. Currently there are almost weekly releases changing things, which would just cause too much of an overhead for us.

### DIFF
--- a/CHANGELOG.next-release.md
+++ b/CHANGELOG.next-release.md
@@ -1,2 +1,2 @@
-* Add support for OpenAI client 0.14+ - #531
+* Add support for OpenAI client 0.14 to 0.31 - #531, #564
 * Add support for OpenAI developer messages and raise minimum supported version to 0.8.0 - #539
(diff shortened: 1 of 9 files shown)


SylvainJuge (approved):
LGTM, are those methods already covered by existing tests ?

Also, not directly related but should we think about adding the base version of the target library in the instrumentation module name to align with upstream ? I vaguely remember we already talked about it but don't recall what the conclusion was.

JonasKunz:
> Also, not directly related but should we think about adding the base version of the target library in the instrumentation module name to align with upstream ? I vaguely remember we already talked about it but don't recall what the conclusion was.

Wasn't this just about the package names? We added the supported base versions there.

> are those methods already covered by existing tests

Our instrumented methods are, yes. The ones we don't instrument are not, but the proxying mechanism itself is tested.

codefromthecrypt (approved):
Thanks for the continued drift management!

PR #566, opened 2025-03-06
Fix env variable for enabling micrometer metrics

anuraaga:


JonasKunz (approved):
Wow, had to check three times to spot the difference. Thanks for fixing this!

PR #583, opened 2025-03-18
Change default metric temporality to delta

### LINKED ISSUE #446: Switch to delta-preferred metrics temporality
As discussed with other agent teams, we should switch to delta-preferred temporality.
This is a "potentially breaking change", but it can be reverted using configuration.

Also we should document somewhere that when using the vanilla otel agent, it is highly recommended to put it into delta-preferred temporality mode.

JonasKunz:
Closes #446 .
I've manually verified that this change is backwards-compatible for our existing Kibana Dashboards: The runtime-metrics dashboard currently available when ingesting OTLP via apm-server only uses gauges and up-down-counters, which are not affected by this temporality change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Set `otel.instrumentation.runtime-telemetry.emit-experimental-telemetry` to `fal
 ### Metric Temporality
 
 Elasticsearch and Kibana work best with metrics provided in delta-temporality.
-Therefore, the EDOT Java changes sets the default value of `otel.exporter.otlp.metrics.temporality.preference` to `DELTA`.
+Therefore, the EDOT Java changes the default value of `otel.exporter.otlp.metrics.temporality.preference` to `DELTA`.
 You can override this default if needed, note though that some provided Kibana dashboards will not work correctly in this case.
 
 # License
--- a/README.md
+++ b/README.md
@@ -79,6 +79,13 @@ in the [`span-stacktrace`](https://github.com/open-telemetry/opentelemetry-java-
 Experimental runtime metrics are enabled by default.
 Set `otel.instrumentation.runtime-telemetry.emit-experimental-telemetry` to `false` to disable them.
 
+
+### Metric Temporality
+
+Elasticsearch and Kibana work best with metrics provided in delta-temporality.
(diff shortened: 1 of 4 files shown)


custom/src/test/java/co/elastic/otel/ElasticAutoConfigurationCustomizerProviderTest.java
```diff
@@ -67,4 +67,13 @@ void disableExperimentalRuntimeMetrics() {
     String value = config.get("otel.instrumentation.runtime-telemetry.emit-experimental-telemetry");
     assertThat(value).isEqualTo("false");
   }
+
+  @Test
+  void customizeMetricTemporalityPreference() {
```
SylvainJuge, line 72: [minor] probably worth adding a test to ensure we override the default.

README.md
```diff
@@ -79,6 +79,13 @@ in the [`span-stacktrace`](https://github.com/open-telemetry/opentelemetry-java-
 Experimental runtime metrics are enabled by default.
 Set `otel.instrumentation.runtime-telemetry.emit-experimental-telemetry` to `false` to disable them.
 
+
+### Metric Temporality
+
+Elasticsearch and Kibana work best with metrics provided in delta-temporality.
+Therefore, the EDOT Java changes sets the default value of `otel.exporter.otlp.metrics.temporality.preference` to `DELTA`.
```
SylvainJuge, line 86:
```suggestion
Therefore, the EDOT Java changes the default value of `otel.exporter.otlp.metrics.temporality.preference` to `DELTA`.
```

We also need to document this in https://github.com/elastic/opentelemetry/blob/main/docs/_edot-sdks/java/features.md, while not being a "feature" by itself we need to document that we have a different default here.

SylvainJuge (dismissed):

custom/src/test/java/co/elastic/otel/ElasticAutoConfigurationCustomizerProviderTest.java
JonasKunz: Fixed in [8790acb](https://github.com/elastic/elastic-otel-java/pull/583/commits/8790acb2a44a26b200e2eb70f415161dac86b0f3)

SylvainJuge (approved):

PR #589, opened 2025-03-21
add runtime attach

SylvainJuge:
## implementation checklist

- [x] bootstrap new module and reuse contrib implementattion
- [x] validate if adding a new custom entry point is the right approach: the main benefit is that it prevents leaking the contrib repo API to be user-facing: OK because:
  - providing custom entry point and a single artifact is simpler for end-user: reusing contrib + asking to configure it may lead to errors
  - allows to provide an artifact whose version is aligned with the embedded agent version
  - very little code involved and implementation is almost trivial
  - provides flexibility on the entry point, for example allowing to execute things very early or to having the ability to provide alternative entry points if needed.
- [x] add integration test


runtime-attach/src/main/java/co/elastic/otel/agent/attach/RuntimeAttach.java
```diff
@@ -0,0 +1,31 @@
...
+ * under the License.
+ */
+package co.elastic.otel.agent.attach;
+
+import io.opentelemetry.contrib.attach.core.CoreRuntimeAttach;
+
+public class RuntimeAttach {
+
+  public static void attachJavaagentToCurrentJvm() {
+    CoreRuntimeAttach distroRuntimeAttach = new CoreRuntimeAttach("/otel-agent.jar");
```
JonasKunz, line 26: Can we use any name for the embedded jar? If yes, I'd prefer `edot-agent.jar`

JonasKunz (dismissed):

runtime-attach/src/main/java/co/elastic/otel/agent/attach/RuntimeAttach.java
SylvainJuge: Yes, we definitely can, I'll change it to `edot-agent.jar` then, the name here comes from copy-pasting in the upstream implementation, but there is no "convention" to make it automatically included.

agent/entrypoint/src/main/java/co/elastic/otel/agent/ElasticAgent.java
```diff
@@ -24,12 +24,34 @@
...
 
+  /**
+   * Entry point for runtime attach
+   *
+   * @param agentArgs agent arguments
+   * @param inst instrumentation
+   */
+  public static void agentmain(String agentArgs, Instrumentation inst) {
+    OpenTelemetryAgent.agentmain(agentArgs, inst);
+  }
```
SylvainJuge, line 45: [for reviewer] there is a different entry point when doing runtime attach, which is not needed otherwise.

runtime-attach/README.md
```diff
@@ -0,0 +1,72 @@
...
+public class MyApplication {
+
+    public static void main(String[] args) {
+        RuntimeAttach.attachJavaagentToCurrentJvm();
+        SpringApplication.run(MyApplication.class, args);
+    }
+}
+```
+
+## JVM Runtime attach warnings
```
JonasKunz, line 58: Very good idea to provide clarifications on the warnings here!

JonasKunz (approved):

PR #596, opened 2025-04-02
enable azure provider by default

SylvainJuge:
Azure resource provider has been added in upstream instrumentation with https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13627

because we enable those providers by default, we can anticipate this addition even if it will only be included in the next upstream release.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -30,8 +30,6 @@ opentelemetryInstrumentationAlphaBom = { group = "io.opentelemetry.instrumentati
 
 opentelemetryProto = { group = "io.opentelemetry.proto", name = "opentelemetry-proto", version.ref = "opentelemetryProto" }
 
-awsContribResources = { group = "io.opentelemetry.contrib", name = "opentelemetry-aws-resources", version.ref = "opentelemetryContribAlpha" }
-gcpContribResources = { group = "io.opentelemetry.contrib", name = "opentelemetry-gcp-resources", version.ref = "opentelemetryContribAlpha" }
 contribResources = { group = "io.opentelemetry.contrib", name = "opentelemetry-resource-providers", version.ref = "opentelemetryContribAlpha" }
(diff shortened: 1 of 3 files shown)


codefromthecrypt (approved):
impact is yet another HTTP request at startup I guess? e.g. if you use none of these and don't set a property to disable, you'll get a 3 types of remote resource requests instead of 2?

And that we accept that due to the policy of doing what upstream does. Correct?

Just saying aloud impact, which may be misunderstood.

SylvainJuge:
Good point, an extra HTTP request is made, but not always:
- most of the azure resource providers are only executed when some environment variables are set
- the one that does an http call is executed last, as we can see in [AzureVmResourceProvider](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/c19a97334afcbc43a9cd07eff4c8f2a3cdb121dc/azure-resources/src/main/java/io/opentelemetry/contrib/azure/resource/AzureVmResourceProvider.java) implementation.
- in the case where none is relevant, the "expensive" one will be systematically executed, which is similar to what the EC2 resource provider does for AWS, as there is no easy and fast way to do that, there are "hack-ish" ways using `dmidecode` but those require root privileges so an http request isn't that bad after all.

So yes, it might make the app startup slightly slower, at the expense of a better out of the box experience. If that becomes an issue we can provide a dedicated documentation to optimize application overhead and startup time.

jackshirazi:
Let's add a startup time section in the doc, we know from experience that it's important for some customers

SylvainJuge:
@jackshirazi I've opened https://github.com/elastic/opentelemetry/pull/193 for that, it anticipates the availability of the feature in EDOT Java, which in fact depends on :
- instrumentation java next release (likely due within about a week)
- update the upstream dependency in this repo (should be an automatic dependency update PR)
- do a release of EDOT Java

PR #607, opened 2025-04-14
Add support for openAI client 1.1+, drop support for older versions

JonasKunz:
The first stable OpenAI Java client (1.0.0) was released a few days ago.
This PR adds support for 1.1+ and drops support for all older versions.

1.0 seems to have moved a lot of classes around causing breaking changes, but looks like nothing functionally changed.
I didn't touch the logic and all tests are green.

I decided to only support 1.1+, as the way to extract the `baseUrl` from the client builder changed and 1.1 was released only a few hours after 1.0. Therefore I don't think it's worth adding a special case / separate module to support 1.0.


codefromthecrypt (dismissed):
Agree to "[draw a line in the sand](https://en.wikipedia.org/wiki/Line_in_the_sand)"

jackshirazi (approved):

PR #610, opened 2025-04-15
Workaround for temporality configuration not being possible

JonasKunz:
Just after releasing, I discovered that the temporality setting `otel.exporter.otlp.metrics.temporality.preference` had no more effect. The temporality would always be cumulative, independent of what is configured. Even our changed default to delta did not take effect.

This was introduced in #593 due to a bug in the `toBuilder()` implementation in the exporters:
https://github.com/open-telemetry/opentelemetry-java/issues/7276

I've added more integrationish test reproducing the problem and added a workaround. I've also manually verified the fix.

### DIFF
--- a/custom/src/main/java/co/elastic/otel/ElasticUserAgentHeader.java
+++ b/custom/src/main/java/co/elastic/otel/ElasticUserAgentHeader.java
@@ -47,12 +47,23 @@ public class ElasticUserAgentHeader {
   }
 
   public static MetricExporter configureIfPossible(MetricExporter metricExporter) {
+    // TODO remove workaround for https://github.com/open-telemetry/opentelemetry-java/issues/7276
+    // once fixed, setAggregationTemporalitySelector and setDefaultAggregationSelector can then be
+    // removed
     if (metricExporter instanceof OtlpGrpcMetricExporter) {
       return ((OtlpGrpcMetricExporter) metricExporter)
-          .toBuilder().addHeader(HEADER_NAME, GRPC_VALUE).build();
+          .toBuilder()
+              .setAggregationTemporalitySelector(metricExporter)
+              .setDefaultAggregationSelector(metricExporter)
+              .addHeader(HEADER_NAME, GRPC_VALUE)
+              .build();
     } else if (metricExporter instanceof OtlpHttpMetricExporter) {
       return ((OtlpHttpMetricExporter) metricExporter)
-          .toBuilder().addHeader(HEADER_NAME, HTTP_VALUE).build();
+          .toBuilder()
+              .setAggregationTemporalitySelector(metricExporter)
+              .setDefaultAggregationSelector(metricExporter)
+              .addHeader(HEADER_NAME, HTTP_VALUE)
+              .build();
     }
     return metricExporter;
   }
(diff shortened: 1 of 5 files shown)


custom/src/test/java/co/elastic/otel/ElasticAutoConfigurationCustomizerProviderTest.java
```diff
@@ -88,6 +95,29 @@ void customizeMetricTemporalityPreference() {
     assertThat(value).isEqualTo("LOWMEMORY");
   }
 
+  @Test
+  void verifyEffectiveDefaultTemporality() {
```
codefromthecrypt, line 99: optional: since the E in EDOT is Elastic, worth a javadoc comment mentioning that these defaults are tuned to what EDOT collector supports? (or which are not)

JonasKunz: I've renamed the test method to clarify that we override the default.
For the reasoning on why we do this, there's a [section in the docs](https://elastic.github.io/opentelemetry/edot-sdks/java/features.html#metric-temporality). We don't do this just for histograms, but also because handling cumulative counters is still not great, though it's getting better with ES|QL.

custom/src/test/java/co/elastic/otel/ElasticAutoConfigurationCustomizerProviderTest.java
```diff
@@ -96,7 +96,7 @@ void customizeMetricTemporalityPreference() {
   }
 
   @Test
-  void verifyEffectiveDefaultTemporality() {
+  void verifyDefaultTemporalityOverriddenToDelta() {
```
codefromthecrypt, line 99: SGTM

SylvainJuge (dismissed):

custom/src/main/java/co/elastic/otel/ElasticUserAgentHeader.java
```diff
@@ -47,12 +47,23 @@ public static SpanExporter configureIfPossible(SpanExporter spanExporter) {
   }
 
   public static MetricExporter configureIfPossible(MetricExporter metricExporter) {
+    // TODO remove workaround for https://github.com/open-telemetry/opentelemetry-java/issues/7276
+    // once fixed, setAggregationTemporalitySelector and setDefaultAggregationSelector can then be
+    // removed
```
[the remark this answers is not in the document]
JonasKunz: > Also, it would be nice to make explicit in the comment what needs to be removed (calls to setAggregationTemporalitySelector and setDefaultAggregationSelector)

But that's waht the comment already does?

>  // once fixed, setAggregationTemporalitySelector and setDefaultAggregationSelector can then be
>  // removed

I'll add a test that fails when the fix is in

custom/src/test/java/co/elastic/otel/ElasticAutoConfigurationCustomizerProviderTest.java
```diff
@@ -88,6 +95,29 @@ void customizeMetricTemporalityPreference() {
...
+    try (OpenTelemetrySdk sdk =
+        AutoConfiguredOpenTelemetrySdk.builder().build().getOpenTelemetrySdk()) {
+      List<MetricExporter> metricExporters =
+          OtelReflectionUtils.extractMetricExporters(sdk.getSdkMeterProvider());
+
+      assertThat(metricExporters).hasSize(1);
+
+      MetricExporter otlp = metricExporters.get(0);
+      assertThat(otlp.getAggregationTemporality(InstrumentType.COUNTER))
+          .isEqualTo(AggregationTemporality.DELTA);
```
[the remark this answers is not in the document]
JonasKunz: Yes, gauges have no temporality

SylvainJuge (approved):

PR #613, opened 2025-04-16
dynamic log level

### LINKED ISSUE #465: [Config] Option to change Agent log level dynamically
For implementation, this will be about changing the log level dynamically in the current instrumentation, which might require some changes upstream or in the SDK to be effective.

A follow-up part will be to control this new feature through the opamp client implementation to finally allow to control this setting remotely.

SylvainJuge:
Fixes #465 

- uses log4j2 as implementation, which provides ability to set level at runtime
- redirects agent internal logger to slf4j, which then delegates to log4j
- this only provides a way to set the log level from within the agent, it does not yet allow to set the log level from user-controlled interface (for example remotely through opamp or a local configuration file).
- introduces `elastic.otel.javaagent.log.level` config option to set the log level using [log4j levels](https://logging.apache.org/log4j/2.x/manual/customloglevels.html) it provides the ability to use `OFF` option to completely disable agent logging.


buildSrc/src/main/kotlin/elastic-otel.agent-packaging-conventions.gradle.kts
```diff
@@ -182,6 +182,9 @@ tasks {
     dependsOn(isolateJavaagentLibs)
     configurations = listOf(bootstrapLibs, upstreamAgent)
 
+    // exclude slf4j-simple from the shadow jar as we use log4j2-slf4j with internal-logging instead
+    exclude("inst/io/opentelemetry/javaagent/slf4j/simple/**")
```
jackshirazi, line 186: What happens if OTEL_JAVAAGENT_LOGGING is set to simple?

gradle/libs.versions.toml
```diff
@@ -4,6 +4,7 @@ jib = "3.4.5"
 spotless = "7.0.3"
 junit = "5.12.2"
 autoservice = "1.1.1"
+log4j2 = "2.24.3"
```
jackshirazi, line 7: do we really want to start explicitly depending on log4j? Any reason not to use j.u.l?

SylvainJuge: JUL static initialization is something that tends to create issues (for example with Wildfly/Jboss), so we'd better avoid it.

Also log4j provides ability to change log level at runtime very easily, so I thought it would be simpler.

buildSrc/src/main/kotlin/elastic-otel.agent-packaging-conventions.gradle.kts
SylvainJuge: Because the implementation is not present, a noop internal logger is used. I think we could maybe transparently convert `simple` with `elastic` in the case someone is taking default configuration and making it explicit to prevent unexpected changes in defaults.

gradle/libs.versions.toml
jackshirazi: Yes, I saw the agent itself uses j.u.l and works around the initialization issues by having it's own Log manager. If we tried to switch the upstream agent from using simple (which can't change dynamically) to another logger, then this would work upstream too. But it may be too much work

SylvainJuge: I agree, for this I took inspiration from MS ApplicationInsights that delegates to logback instead of log4j. We could maybe challenge this again upstream and suggest to use either slf4j or an internal logging API in the instrumentation to avoid having to deal with JUL at all in the agent, but that would likely be too much work for little benefit.

JonasKunz (dismissed):
LGTM, though I didn't test the code myself

internal-logging/src/main/java/co/elastic/otel/logging/AgentLog.java
```diff
@@ -0,0 +1,69 @@
...
+  private static final String PATTERN = "%d{DEFAULT} [%t] %-5level %logger{36} - %msg{nolookups}%n";
+
+  // logger is an empty string
+  private static final String ROOT_LOGGER_NAME = "";
+
+  private AgentLog() {}
+
+  public static void init() {
+
+    ConfigurationBuilder<BuiltConfiguration> conf =
```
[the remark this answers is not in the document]
SylvainJuge: That's a very good question, I need to investigate a bit to see if there is a simple and easy way to provide a 100% programmatic setup that can't be influenced by system properties or environment variables.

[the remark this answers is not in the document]
SylvainJuge: You were right to worry about this, the programmatic configuration is also influenced by the global JVM system properties and environment variables (which we can't even hide during initialization).

While there are still no automated tests for this, I've managed to trigger the issue manually by using `LOG4J_DISABLE_JMX=false` environment variable which exposes log4j to JMX.

d6d14e4 commit fixes that by excluding the SPI implementations that are included by default.

buildSrc/src/main/kotlin/elastic-otel.agent-packaging-conventions.gradle.kts
SylvainJuge: Trying to set `OTEL_JAVAAGENT_LOGGING=simple` does produce an ugly SLF4J error:

```
SLF4J(E): A service provider failed to instantiate:
io.opentelemetry.javaagent.slf4j.spi.SLF4JServiceProvider: Provider io.opentelemetry.javaagent.slf4j.simple.SimpleServiceProvider not found
SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
```

So I've changed the behavior in https://github.com/elastic/elastic-otel-java/pull/613/commits/0b05c841bb2eb1710310bf23545af3431b491644, now when it's explicitly set to `simple` we still override the provided value and use the `elastic` logger.

JonasKunz (approved):

PR #643, opened 2025-05-19
Workaround for upstream bug triggering nested autoconfiguration

JonasKunz:
Workaround for https://github.com/open-telemetry/opentelemetry-java/issues/7354, should only affect and fix the flaky smoke test since the last upgrade of the upstream SDK version (#641).

### DIFF
--- a/smoke-tests/src/test/java/com/example/javaagent/smoketest/UserAgentHeaderTest.java
+++ b/smoke-tests/src/test/java/com/example/javaagent/smoketest/UserAgentHeaderTest.java
@@ -157,19 +157,19 @@ public class UserAgentHeaderTest {
                 // Traces ands logs are generated via the container health check
                 verify(
                     postRequestedFor(
-                        urlEqualTo(
-                            "/opentelemetry.proto.collector.trace.v1.TraceService/Export"))
+                            urlEqualTo(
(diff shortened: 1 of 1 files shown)


smoke-tests/src/test/java/com/example/javaagent/smoketest/UserAgentHeaderTest.java
```diff
@@ -188,7 +188,7 @@ private GenericContainer<?> runTestApp(Consumer<GenericContainer<?>> customizer)
             .withEnv("OTEL_BSP_SCHEDULE_DELAY", "10")
             .withEnv("OTEL_BLRP_MAX_EXPORT_BATCH", "1")
             .withEnv("OTEL_BLRP_SCHEDULE_DELAY", "10")
-            .withEnv("OTEL_METRIC_EXPORT_INTERVAL", "10")
```
SylvainJuge, line 191: can you add a comment with a link to the issue so we can later refer to this when trying to properly fix the underlying issue ?

JonasKunz: Done in https://github.com/elastic/elastic-otel-java/pull/643/commits/2914b9d4ea356a202601c6e5601c3d2ef5abef97

PR #646, opened 2025-05-20
add basic opamp implementation

jackshirazi:
This is pretty much moving https://github.com/elastic/opamp-java (and a snapshot of the opamp-proto) into this repo, with some minor changes. It's intended as an interim solution as the opamp implementation will be contributed upstream and when done this will be removed and that dependency will be added.

This is the first PR in a series to get to the agent handling central config, I've broken the full implementation into sections to make it more manageable

The new added files are all copied from the opamp repo (there are some minor changes) - I don't think they need in-depth review, just a scan that there's nothing silly in them. The three modified files  (`settings.gradle.kts`, `custom/build.gradle.kts`, `buildscripts/allowed-licenses.json`) are for the build - please spend a bit more time checking these changes to ensure they are fine


opamp/src/main/java/co/elastic/opamp/client/connectivity/websocket/OkHttpWebSocket.java
LikeTheSalad, line 1: nit: maybe we can remove all the websocket-related code, as my understanding is that we're not planning to use it here.

opamp/src/main/java/opamp/proto/Anyvalue.java
LikeTheSalad, line 1: It's not a blocker, though I'm curious about the decision to add the generated proto classes into the repo.

LikeTheSalad (approved):

opamp/src/main/java/opamp/proto/Anyvalue.java
jackshirazi: It's a snapshot of the current proto. This whole implementation is interim and we'll switch to the upstream contrib as soon as that's available, in the meantime we'll be on this version of proto and I think that keeps the maintenance to a minimum which is what you want for an interim implementation. I started with putting the proto repo in too, but that's just the same snapshot and this removes one build step that will (likely) never change

opamp/src/main/java/co/elastic/opamp/client/connectivity/websocket/OkHttpWebSocket.java
jackshirazi: I'll leave it for now, since the websocket is actually the default, so maybe we'll need it

PR #647, opened 2025-05-21
refactor deactivated instrumentations to be remembered (not just in test)

jackshirazi:
This was previously only needed in the test loop, but the central config will send a list of instrumentations and this needs to be handled in the dynamic config class the same way, so I'm just moving it across

This is the second PR in a series to get to the agent handling central config, I've broken the full implementation into sections to make it more manageable

### DIFF
--- a/custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
+++ b/custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
@@ -182,7 +182,7 @@ public class DynamicConfiguration {
     final Set<String> instrumentationsToReactivate;
     final Set<String> instrumentationsToDeactivate;
 
-    public <E> Deactivations(Set<String> deactivateList, Set<String> alreadyDeactivated) {
+    public Deactivations(Set<String> deactivateList, Set<String> alreadyDeactivated) {
       instrumentationsToReactivate = new HashSet<>();
       instrumentationsToDeactivate = new HashSet<>();
       for (String instrumentation : deactivateList) {
--- a/custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
+++ b/custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
@@ -134,8 +134,8 @@ public class DynamicConfiguration {
     reenableTracesFor(ALL_INSTRUMENTATION);
   }
 
-  //okay to synchronize as this should only be called after multi-second intervals and
-  //additionally only called from threads which are not doing anything application blocking
+  // okay to synchronize as this should only be called after multi-second intervals and
+  // additionally only called from threads which are not doing anything application blocking
   public synchronized void deactivateInstrumentations(String deactivateList) {
     if (deactivateList != null && !deactivateList.trim().isEmpty()) {
       // some values in the disable_instrumentations list
--- a/custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
+++ b/custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
@@ -157,7 +157,7 @@ public class DynamicConfiguration {
       Set<String> keySet = alreadyDeactivated.keySet();
       for (String instrumentation : keySet) {
         DynamicConfiguration.getInstance().reenableTracesFor(instrumentation);
-        alreadyDeactivated.remove(instrumentation);
+        keySet.remove(instrumentation);
       }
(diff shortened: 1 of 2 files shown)


custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
```diff
@@ -131,6 +134,44 @@ public void stopDisablingAllTraces() {
     reenableTracesFor(ALL_INSTRUMENTATION);
   }
 
+  // okay to synchronize as this should only be called after multi-second intervals and
+  // additionally only called from threads which are not doing anything application blocking
+  public synchronized void deactivateInstrumentations(String deactivateList) {
+    if (deactivateList != null && !deactivateList.trim().isEmpty()) {
+      // some values in the disable_instrumentations list
+      Set<String> toBeEnabled = null;
```
SylvainJuge, line 142: [minor] making this non-nullable would remove a couple of null-checks below

custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
```diff
@@ -131,6 +134,44 @@ public void stopDisablingAllTraces() {
     reenableTracesFor(ALL_INSTRUMENTATION);
   }
 
+  // okay to synchronize as this should only be called after multi-second intervals and
+  // additionally only called from threads which are not doing anything application blocking
+  public synchronized void deactivateInstrumentations(String deactivateList) {
```
SylvainJuge, line 139: [minor] could we delegate the implementation to a static package-private method and test it so that it would just compute the two `to activate` and `to disable` sets and update the `alreadyDeactivated` collection only in the public non-static method ?

jackshirazi: It's already being tested via the integration smoke test which uses the active DynamicConfigurationPropertyChecker - which in turn needs this to be working correctly

I think splitting it how you suggest would make it more complicated, but I'll play and see

custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
jackshirazi: the null status is a specific state where there is something already deactivated. If there is nothing deactivated the path is different. Essentially making this non-null would then require another boolean variable

jackshirazi: not relevant with the changed flow

custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
jackshirazi: updated

custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
```diff
@@ -131,6 +136,81 @@ public void stopDisablingAllTraces() {
...
+      }
+    }
+    return theSet;
+  }
+
+  public static class Deactivations {
+    final Set<String> instrumentationsToReactivate;
+    final Set<String> instrumentationsToDeactivate;
+
+    public <E> Deactivations(Set<String> deactivateList, Set<String> alreadyDeactivated) {
```
SylvainJuge, line 185:
```suggestion
    public Deactivations(Set<String> deactivateList, Set<String> alreadyDeactivated) {
```

custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
```diff
@@ -131,6 +136,81 @@ public void stopDisablingAllTraces() {
...
+      return;
+    }
+    cachedDeactivationList = deactivateList;
+
+    // Algorithm:
+    // 1. If the list is empty, then everything alreadyDeactivated needs to be re-activated
+    // 2. Otherwise
+    // 2a. everything in both deactivateList and alreadyDeactivated is ignored (already deactivated)
+    // 2b. everything in deactivateList not in alreadyDeactivated is deactivated
+    // 2c. everything in alreadyDeactivated not in deactivateList is re-eactivated
```
SylvainJuge, line 154:
```suggestion
    // 2c. everything in alreadyDeactivated not in deactivateList is re-activated
```

SylvainJuge (dismissed):

SylvainJuge (approved):

PR #650, opened 2025-05-21
Align dynamic config API to kibana send* config options

jackshirazi:
We decided to provide per signal send_* config options in central config instead of just one global send, so we need those in teh dynamic config API


custom/src/main/java/co/elastic/otel/dynamicconfig/DynamicConfiguration.java
```diff
@@ -84,6 +84,54 @@ private void initSendingStates() {
     }
   }
 
+  /** Can be executed repeatedly even if sending is currently stopped */
+  public void stopSendingSpans() {
```
JonasKunz, line 88: Is there a particular reason to having start/stop methods instead of just a `setSendingSpans(boolean)` method?

PR #656, opened 2025-05-27
connect opamp to dynamic config

jackshirazi:
Mainly adding and activating a new CentralConfig class, plus a little refactoring to align method names to option names. No test of the CentralConfig class in this PR, I'll do that in a followup. This is almost the last PR in the sequence to get central config working and integrated to dynamic config, the only thing left is that test and some additional work around property/env var names for the new functionality

### DIFF
--- a/custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
+++ b/custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
@@ -1,0 +1,271 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.otel.dynamicconfig;
+
+import co.elastic.opamp.client.CentralConfigurationManager;
+import co.elastic.opamp.client.CentralConfigurationProcessor;
+import co.elastic.otel.logging.AgentLog;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder;
+import java.text.MessageFormat;
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Logger;
+
+public class CentralConfig {
+  private static final Logger logger = Logger.getLogger(CentralConfig.class.getName());
+
+  static {
+    DynamicConfigurationPropertyChecker.startCheckerThread();
+  }
+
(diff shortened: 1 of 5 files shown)


custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
```diff
@@ -0,0 +1,271 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Logger;
+
+public class CentralConfig {
+  private static final Logger logger = Logger.getLogger(CentralConfig.class.getName());
+
+  static {
+    DynamicConfigurationPropertyChecker.startCheckerThread();
```
JonasKunz, line 38: Do we always want to start this or only if opamp is actually enabled?
If the latter, we should probably move this to the `init` method instead?

custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  public static void init(SdkTracerProviderBuilder providerBuilder, ConfigProperties properties) {
+    // TODO flip default when EDOT collector supports op amp
+    boolean startOpAmp = properties.getBoolean("elastic.otel.opamp.start", false);
+    if (!startOpAmp) {
+      return;
+    }
+    String serviceName = getServiceName(properties);
```
JonasKunz, line 47: This won't take service names derived from `ResourceProvider`s into account, will it?
If not, we should probably add a TODO here?

custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
```diff
@@ -0,0 +1,271 @@
...
+
+    static {
+      ConfigOption option;
+      configNameToConfig.put((option = new SendLogs()).getConfigName(), option);
+      configNameToConfig.put((option = new SendMetrics()).getConfigName(), option);
+      configNameToConfig.put((option = new SendTraces()).getConfigName(), option);
+      configNameToConfig.put(
+          (option = new DeactivateAllInstrumentations()).getConfigName(), option);
+      configNameToConfig.put((option = new DeactivateInstrumentations()).getConfigName(), option);
+      configNameToConfig.put((option = new LoggingLevel()).getConfigName(), option);
```
JonasKunz, line 112: Nit: A matter of tase but I prefer the functional approach here:
```
configNameToConfig = Stream.of(
    new SendLogs(),
    new SendMetrics(),
    new SendTraces(),
    new DeactivateAllInstrumentations(),
    new DeactivateInstrumentations(),
    new LoggingLevel()
).collect(Collectors.toMap(ConfigOption::getConfigName, option -> option))
```

JonasKunz (approved):

custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
jackshirazi: always, it's primarily for testing though can be used in production

custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
jackshirazi: that is nicer, I agree

custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
JonasKunz: I was asking because `getServiceName(properties)` only reads from the `ConfigurationProperties` and not potential changes from `ResourceProviders` which don't write those, but directly set the `Resource`, e.g. [AppServiceNameProvider](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/f4ff5f4b02df0b1572233ca9bfdd124489cd3b53/resource-providers/src/main/java/io/opentelemetry/contrib/resourceproviders/AppServerServiceNameProvider.java#L24).

But maybe there is some magic behind the curtains making this work? So if you tested it this is fine, otherwise we should add a TODO

JonasKunz (approved):

PR #668, opened 2025-06-03
only elastic.otel.opamp.endpoint or env var are required

jackshirazi:
Aligning to decision that OpAMP is off by default but is enabled if elastic.otel.opamp.endpoint(or the equivalent env var) is defined, and that the "/v1/opamp" path is optional, so will be auto-added if not there

### DIFF
--- a/custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
+++ b/custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
@@ -40,16 +40,11 @@ public class CentralConfig {
   }
 
   public static void init(SdkTracerProviderBuilder providerBuilder, ConfigProperties properties) {
-    // TODO flip default when EDOT collector supports op amp
-    boolean startOpAmp = properties.getBoolean("elastic.otel.opamp.start", false);
-    if (!startOpAmp) {
+    String endpoint = properties.getString("elastic.otel.opamp.endpoint");
+    logger.info("Enabling OpAMP as endpoint is defined: " + endpoint);
+    if (endpoint == null || endpoint.isEmpty()) {
       return;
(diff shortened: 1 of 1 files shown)


custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
```diff
@@ -40,30 +40,27 @@ public class CentralConfig {
   }
 
   public static void init(SdkTracerProviderBuilder providerBuilder, ConfigProperties properties) {
-    // TODO flip default when EDOT collector supports op amp
-    boolean startOpAmp = properties.getBoolean("elastic.otel.opamp.start", false);
-    if (!startOpAmp) {
+    String endpoint = properties.getString("elastic.otel.opamp.endpoint");
+    if (endpoint == null || endpoint.isEmpty()) {
+      logger.fine("OpAMP is not enabled");
```
SylvainJuge, line 45:
```suggestion
      logger.fine("OpAMP is disabled");
```
The negation is likely to not be read and people might read it as "enabled" when it's "not enabled.

custom/src/main/java/co/elastic/otel/dynamicconfig/CentralConfig.java
```diff
@@ -95,7 +92,7 @@ private static String getServiceName(ConfigProperties properties) {
         return serviceName;
       }
     }
-    return "unknown_service"; // Specified default
+    return "unknown_service:java"; // Specified default
```
SylvainJuge, line 95: Do we have a link to the specification here ? If yes, then maybe adding it as a comment would be helpful for future reference.

SylvainJuge (approved):

PR #672, opened 2025-06-04
Update README.md links

jackshirazi:
The "X.html" pages don't resolve, they need ".html" stripped except for index (prob cos historically index pages are that)


theletterf (approved):
Thanks!

PR #679, opened 2025-06-09
[ci] simplify usage of is-member-of

SylvainJuge:
We should not have to ignore bots by name as we found when working on https://github.com/elastic/oblt-actions/pull/290.


v1v (approved):
Thanks!

PR #702, opened 2025-06-23
Add docs CI files

theletterf:
These two actions will allow us to preview and publish the release notes and other reference docs.


karenzone (approved):
LGTM

PR #704, opened 2025-06-24
implement spans debug logging

SylvainJuge:
With the upstream otel agent, when the `otel.javaagent.debug` option is set to `true`, an additional span logging exporter is added when there is none, that means spans start/and are logged at `INFO` level when `otel.javaagent.debug` is set to `true`.

Our current implementation of dynamic log level does currently replicates this behavior, which means that when the dynamic log level is set to `DEBUG` the resulting behavior is not exactly equivalent to setting `otel.javaagent.debug=true` with the upstream agent.

This PR replicates this behavior by doing the following:
- when an logging exporter is already configured, we let the configured one log the spans, which means span logging will happen whenever log level is `INFO` or above.
- when no logging exporter is configured, we add a dedicated `SpanExporter` that can be controlled at runtime, then when the log level is set to `DEBUG` or above we can then control this extra span exporter at runtime when needed.
- when the upstream debug config option is used, use the same plain text output as the upstream agent
- when setting our own logger level to `debug`, use the json output for easier log parsing.

Feature documentation is covered in https://github.com/elastic/opentelemetry/pull/315

### DIFF
--- a/internal-logging/src/main/java/co/elastic/otel/logging/AgentLog.java
+++ b/internal-logging/src/main/java/co/elastic/otel/logging/AgentLog.java
@@ -153,7 +153,7 @@ public class AgentLog {
     }
 
     // when debugging the upstream otel agent configures an extra debug exporter
-    if(debugLogSpanExporter != null) {
+    if (debugLogSpanExporter != null) {
       debugLogSpanExporter.setEnabled(isDebug);
     }
   }
--- a/internal-logging/src/main/java/co/elastic/otel/logging/AgentLog.java
+++ b/internal-logging/src/main/java/co/elastic/otel/logging/AgentLog.java
@@ -71,15 +71,12 @@ public class AgentLog {
   public static void addSpanLoggingIfRequired(
       SdkTracerProviderBuilder providerBuilder, ConfigProperties config) {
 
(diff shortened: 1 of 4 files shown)


internal-logging/src/main/java/co/elastic/otel/logging/AgentLog.java
```diff
@@ -47,6 +82,27 @@ public static void init() {
...
+
+    // Replicate behavior of the upstream agent: span logging exporter is automatically added when
+    // not already present when debugging. When logging exporter has been explicitly configured,
+    // spans logging will be done by the explicitly configured logging exporter instance.
+
+    logPlainText = config.getBoolean(OTEL_JAVAAGENT_DEBUG, false);
+    String exporterName = logPlainText ? "logging" : "otlp-logging";
+
+    boolean loggingExporterNotAlreadyConfigured =
+        !config.getList("otel.traces.exporter", emptyList()).contains(exporterName);
```
jackshirazi, line 96: if debug is off, `exporterName` is `otlp-logging`. Then if `otel.traces.exporter` includes `logging` (not `otlp-logging`) then the otlp logger is added, so we end up with both loggers. That actually seems fine to me, but I wanted to confirm that's what you intended

jackshirazi (approved):

PR #789, opened 2025-09-11
Opamp migration

### LINKED ISSUE #788: Use upstream OpAMP client
Replacing our internal opamp client impl with the one from the [upstream repo](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/opamp-client).

LikeTheSalad:
Closes #788


jackshirazi:
Upstream is [missing exponential backoff](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1956) which we currently have available in the agent. I think it's okay to miss out that feature for now, but we need to prioritize it upstream

jackshirazi (approved):

LikeTheSalad:
> ah, I test with an unavailable port, so connection refused

Got it. Still, I took the time and added a test case for exponential backoff in `OpampManagerTest` just to make sure it works well even after potential upstream changes in the future, as it's still incubating.

PR #818, opened 2025-10-01
Add 9.2 supported dynamic config options

jackshirazi:
Contributes to https://github.com/elastic/opentelemetry-dev/issues/992

### DIFF
--- a/docs/reference/edot-java/configuration.md
+++ b/docs/reference/edot-java/configuration.md
@@ -119,10 +119,10 @@ You can modify the following settings for EDOT Java through APM Agent Central Co
 | Turn off all instrumentations | `deactivate_all_instrumentations` | Dynamic |
 | Send traces | `send_traces` | Dynamic |
 | Send metrics | `send_metrics` | Dynamic |
-| Send logs | `send_logs` | Dynamic {applies_to}`edot_java: preview 1.6.0` |
+| Send logs | `send_logs` | Dynamic  |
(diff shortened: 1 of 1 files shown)


docs/reference/edot-java/configuration.md
```diff
@@ -119,7 +119,10 @@ You can modify the following settings for EDOT Java through APM Agent Central Co
 | Turn off all instrumentations | `deactivate_all_instrumentations` | Dynamic |
 | Send traces | `send_traces` | Dynamic |
 | Send metrics | `send_metrics` | Dynamic |
-| Send logs | `send_logs` | Dynamic |
+| Send logs | `send_logs` | Dynamic  |
+| OpAMP polling interval | `opamp_polling_interval` | Dynamic {applies_to}`edot_java: preview 1.6.0`|
+| Sampling rate | `sampling_rate` | Dynamic {applies_to}`edot_java: preview 1.6.0`|
+| Turn on/off inferred spans | `infer_spans` | Dynamic {applies_to}`edot_java: preview 1.6.1`|
```
SylvainJuge, line 125: [minor] we often refer to the feature as "inferred spans", and the options for the classic agent already align with that, is `infer_spans` the new cross-platform config option we settled with ?

jackshirazi: We are giving the edot agent options different names from the classic agent. The feature is still called Inferred spans, but the option is `infer_spans` which is grammatically correct. These are really internal option names, as they are not exposed in any public API so we could theoretically change them in the future though I don't see the need.

theletterf (approved):

Issue #692, opened 2025-06-17
Automate release note markdown generation


JonasKunz:
#673 adds the release notes as markdown for the central documentation. We should copy the [approach](https://github.com/elastic/apm-agent-java/blob/main/.ci/ReleaseChangelog.java) from elastic/apm-agent-java to have those automatically updated on release.

PR #824, opened 2025-10-06
release changelog automation

SylvainJuge:
Fixes https://github.com/elastic/elastic-otel-java/issues/692

This is a first step to improve release automation, there are definitely rough edges and future improvements will be necessary. In particular, it adds an `DEPENDENCIES-NOTES` section at runtime to include the dependencies versions. In the future maybe managing everything within gradle would be cleaner and more reliable.

We can test it locally for the release 1.6.0 in the current state of main by running the following commands:

```bash
echo '<!--DEPENDENCIES-NOTES-START-->' >> CHANGELOG.next-release.md
echo -e "This release is based on the following upstream versions:\n\n" >> CHANGELOG.next-release.md
./gradlew -q printUpstreamDependenciesMarkdown >> CHANGELOG.next-release.md
echo '<!--DEPENDENCIES-NOTES-END-->' >> CHANGELOG.next-release.md

java \
.ci/ReleaseChangelog.java \
CHANGELOG.next-release.md \
docs/release-notes 1.6.0

```

### DIFF
--- a/.ci/ReleaseChangelog.java
+++ b/.ci/ReleaseChangelog.java
@@ -35,216 +35,234 @@ import java.util.regex.Pattern;
 
 public class ReleaseChangelog {
 
-    public static void main(String[] args) throws IOException {
-        if (args.length != 3) {
-            System.out.println("Expected exactly three arguments: <ChangelogFile> <ReleaseNotesPath> <VersionToRelease>");
-            System.exit(-1);
+  public static void main(String[] args) throws IOException {
+    if (args.length != 3) {
+      System.out.println(
+          "Expected exactly three arguments: <ChangelogFile> <ReleaseNotesPath> <VersionToRelease>");
+      System.exit(-1);
+    }
+    Path nextChangelogFile = Paths.get(args[0]);
+    Path releaseNotesDir = Paths.get(args[1]);
+    Path releaseNotesFile = releaseNotesDir.resolve("index.md");
+    Path deprecationsFile = releaseNotesDir.resolve("deprecations.md");
+    Path breakingChangesFile = releaseNotesDir.resolve("breaking-changes.md");
+    VersionNumber version = VersionNumber.parse(args[2].trim());
+
+    Lines nextChangelogLines = new Lines(
+        Files.readAllLines(nextChangelogFile, StandardCharsets.UTF_8));
+    Lines fixes = nextChangelogLines.cutLinesBetween("<!--FIXES-START-->", "<!--FIXES-END-->");
+    Lines enhancements = nextChangelogLines.cutLinesBetween("<!--ENHANCEMENTS-START-->",
+        "<!--ENHANCEMENTS-END-->");
+    Lines deprecations = nextChangelogLines.cutLinesBetween("<!--DEPRECATIONS-START-->",
+        "<!--DEPRECATIONS-END-->");
+    Lines breakingChanges = nextChangelogLines.cutLinesBetween("<!--BREAKING-CHANGES-START-->",
+        "<!--BREAKING-CHANGES-END-->");
+
+    var formatter = DateTimeFormatter.ofPattern("LLLL d, yyyy", Locale.ENGLISH);
+    String releaseDateLine = "**Release date:** " + formatter.format(LocalDate.now());
+
+    Lines allReleaseNotes = new Lines(Files.readAllLines(releaseNotesFile, StandardCharsets.UTF_8));
+    int insertBeforeLine = findHeadingOfPreviousVersion(allReleaseNotes, version);
+    allReleaseNotes.insert(generateReleaseNotes(version, releaseDateLine, enhancements, fixes),
+        insertBeforeLine);
+
+    if (!deprecations.isEmpty()) {
+      Lines allDeprecations = new Lines(
+          Files.readAllLines(deprecationsFile, StandardCharsets.UTF_8));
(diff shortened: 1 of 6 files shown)


.ci/ReleaseChangelog.java
```diff
@@ -0,0 +1,295 @@
...
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Locale;
+import java.util.OptionalInt;
+import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ReleaseChangelog {
```
SylvainJuge, line 36: [for reviewer] almost copied as-is from apm-agent-java with minor tweaks to handle the lack of previous version.

.ci/ReleaseChangelog.java
```diff
@@ -0,0 +1,295 @@
...
+    Lines fixes = nextChangelogLines.cutLinesBetween("<!--FIXES-START-->", "<!--FIXES-END-->");
+    Lines enhancements = nextChangelogLines.cutLinesBetween("<!--ENHANCEMENTS-START-->",
+        "<!--ENHANCEMENTS-END-->");
+    Lines deprecations = nextChangelogLines.cutLinesBetween("<!--DEPRECATIONS-START-->",
+        "<!--DEPRECATIONS-END-->");
+    Lines breakingChanges = nextChangelogLines.cutLinesBetween("<!--BREAKING-CHANGES-START-->",
+        "<!--BREAKING-CHANGES-END-->");
+
+    Lines dependenciesNotes = nextChangelogLines.cutLinesBetween("<!--DEPENDENCIES-NOTES-START-->",
+        "<!--DEPENDENCIES-NOTES-END-->");
```
SylvainJuge, line 62: [for reviewer] this section does not exists in the next release changelog template, we have to append it manually at runtime. This is due to the fact that we can't easily invoke the gradle task from here, doing everything from gradle in the future would simplify this.

.github/workflows/release-step-3.yml
```diff
@@ -218,6 +218,8 @@ jobs:
           command: ""
       - name: Print Release Notes
         id: print_release_notes
+        # note: release notes here will be copied as-is from 'CHANGELOG.next-release.md'
+        # the 'pre-post-release' workflow executed after this will reset contents of 'CHANGELOG.next-release.md'
```
SylvainJuge, line 222: [for reviewer] here the github release changelog needs to be in pure markdown, so we can't append our custom markdown from breaking changes.

docs/release-notes/index.md
```diff
@@ -27,7 +27,7 @@ To check for security updates, go to [Security announcements for the Elastic sta
 % ### Fixes [edot-java-X.X.X-fixes]
 % *
 
-# 1.5.0 [edot-java-1.5.0-release-notes]
+## 1.5.0 [edot-java-1.5.0-release-notes]
```
SylvainJuge, line 30: [for reviewer] this made the changelog generation not properly detect the previous version.

theletterf:
@jackshirazi @SylvainJuge When that happens, we'll have to use https://github.com/elastic/elastic-otel-java/blob/main/docs/release-notes/breaking-changes.md

v1v (approved):

.github/workflows/pre-post-release.yml
```diff
@@ -100,21 +95,14 @@ jobs:
...
-
-      - name: Clear next release changelog (post release)
+      - name: Generate documentation changelog (post release)
         if: inputs.phase == 'post'
         run: |
-          echo '' > CHANGELOG.next-release.md
+          echo '<!--DEPENDENCIES-NOTES-START-->' >> CHANGELOG.next-release.md
+          echo -e "This release is based on the following upstream versions:\n\n" >> CHANGELOG.next-release.md
+          ./gradlew -q printUpstreamDependenciesMarkdown >> CHANGELOG.next-release.md
+          echo '<!--DEPENDENCIES-NOTES-END-->' >> CHANGELOG.next-release.md
```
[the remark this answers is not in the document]
SylvainJuge: `printUpstreamDependenciesMarkdown` is also used to generate the github release markdown, which should not include those, again this is a shortcut and I'll create an issue with all the things we could improve here.

SylvainJuge:
> Interesting! For now, we don't use have Known issues / Breaking changes pages for EDOT Java, but those could eventually be filled out. Does this PR account for that possibility? Does it produce a PR we can review before merging? And, if we edit the PR, would subsequent runs override the content?

This PR just tries to fills what we currently have in the changelog, I think that having a more structured approach later is needed: for example a yaml description of changes that we can output in multiple markdown variants, for now it's mostly copy-pasta markdown.

> A more general approach for release notes automation is in the works (+CC @alexandra5000 @lcawl @colleenmcginnis ). While I think it's totally fine to implement some automation right now, I guess we might want to align in the future.

Is there already an existing issue to track progress and requirements here ?

On our side we need:
- a way to describe the changes in a structured way, yaml seems a good fit
- having the ability to generate/update markdown for changelog, also for github release notes publication
- having the ability to get the upstream/embedded dependencies versions as 99% of EDOT is actually from upstream repositories (it is currently provided through gradle).

For example, the way to handle changes in otel semconv is quite interesting with a tool called `chloggen` where each PR has its own yaml from a template in a folder and they are removed on release: https://github.com/open-telemetry/semantic-conventions/tree/main/.chloggen, hence also preventing conflicts like we often have with a single changelog file.

PR #840, opened 2025-10-16
stop windows build failures

jackshirazi:


universal-profiling-integration/src/test/java/co/elastic/otel/UniversalProfilingProcessorAutoConfigTest.java
```diff
@@ -34,8 +34,11 @@
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
 import org.junit.jupiter.api.io.TempDir;
 
+@DisabledOnOs(OS.WINDOWS)
 public class UniversalProfilingProcessorAutoConfigTest {
```
LikeTheSalad: I'm not familiarised with this feature, so maybe it's not worth checking it for Windows, but if that's not the case, do you think it's worth creating an issue to fix the tests for Windows later?

jackshirazi: it can't run on windows

PR #841, opened 2025-10-17
make edot sampler default instead of forcing it on

anuraaga:
This updates to the proposal for sampler behavior for EDOT.

/cc @trentm


jackshirazi (approved):
Thank you for this improvement!

SylvainJuge (approved):

Issue #826, opened 2025-10-06
selectively ignore HTTP spans + configure it with declarative configuration


jackshirazi:
## Is your feature request related to a problem?
Like [transaction_ignore_urls](https://www.elastic.co/docs/reference/apm/agents/java/config-http#config-transaction-ignore-urls), this was [the most requested feature](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1060) in the upstream agent, it's closed as you can use a sampler such as [this](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/samplers)

## Describe the solution you'd like
Add the RuleBasedRoutingSampler into the sampler chain of the agent, and provide an option in central config to add/remove paths

## Describe alternatives you've considered
DIY sampler

## Additional context
The upstream agent has added the RuleBasedRoutingSampler as a dependency in 2.21.0 so as soon as we point at that, look at the feasibility of using it - I haven't checked if you can change it dynamically which may be a blocker

PR #900, opened 2025-12-04
[draft] HTTP user-agent and URL filtering

SylvainJuge:
Relates to #826 

This is a draft implementation: 
- ~~reusing rule based sampler implementation from contrib~~
  - update: switched to "composable sampler" implementation in the SDK (incubating)
- using `ELASTIC_OTEL_EXPERIMENTAL_HTTP_IGNORE_URLS` config option to ignore URLs (regex)
- using `ELASTIC_OTEL_EXPERIMENTAL_HTTP_IGNORE_USER_AGENTS` config option to ignore user-agents (regex)
- runtime configuration not implemented yet

### Next Steps
- [x] switch implementation to use "composable rule based sampler" that is/should be included in the SDK.
- [ ] settle on configuration options names 
- [x] settle on configuration options syntax (regex or wildcards ?)
  - update: using wildcards as it's what is being used in composable rule based sampler + part of the configuration recommendations.
- [x] add support for runtime configuration
  - using `http_ignore_urls` and `http_ignore_user_agents` as keys
- [x] add unit test for filtering on span attributes
- [ ] add end-to-end test, maybe also covering the runtime configuration


SylvainJuge:
one of the next steps would be to switch to "composable rule based sampler" as it should now be available and would help switch to wildcards.

SylvainJuge:
closing as we aim to use declarative config with + rule-based sampler configuration to support this in the future.

PR #951, opened 2026-01-29
Disable explicit cache usage to reduce the cache poisoning vector attack

fr4nc1sc0-r4m0n:
## Summary

After a deep analysis, remove explicit cache usage in release related workflow to reduce the [cache poisoning](https://adnanthekhan.com/2024/05/06/the-monsters-in-your-build-cache-github-actions-cache-poisoning/) vector attack.

### Why

In the context of setting up Java, the `actions/setup-java` is explicitly using the gradle cache format.
If a workflow running untrusted code has access to the cache, it can poison the cache with malicious content. For example, a compromised dependency could inject malicious code into a Golang binary or Docker image, which would then be included in subsequent builds using the same cache key.

This attack is particularly dangerous because the malicious artefact can be signed and distributed as part of an SLSA Level 3 build, leaving no trace in the source code or build logs.

### DIFF
--- a/.github/workflows/release-step-3.yml
+++ b/.github/workflows/release-step-3.yml
@@ -147,6 +147,9 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+        with:
+          # NOTE: disable the cache poisoning vector attack
+          cache-binary: false
 
       - name: Log in to the Elastic Container registry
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle-goal/action.yml
```diff
@@ -28,7 +28,8 @@ runs:
       with:
         java-version-file: ${{ inputs.java-version-file }}
         distribution: ${{ inputs.distribution }}
-        cache: 'gradle'
```
mallendem: is there a reason we're not removing the option? I'm fine with the comment, maybe a link to the issue explaining?

mallendem: Thats fair

Issue #1054, opened 2026-04-16
declarative config : base configuration of EDOT 


SylvainJuge:
This issue aims to track progress on adding support for declarative configuration in EDOT Java, the implementation will be split into multipe PRs.

This list has been initially created by reviewing what is being done in `ElasticAutoConfigurationCustomizerProvider` for automatic configuration.

- [x] override default `telemetry.distro.name` and `telemetry.distro.version` #1050 
- [x] add user-agent customization #1050 
- [ ] add ability to skip server certificate verification
  - there is no way to call `setSslContext` with declarative config, so we have to contribute this upstream
  - moved to a separate issue #1084 
- [x] enable experimental runtime telemetry
  - https://github.com/elastic/elastic-otel-java/pull/1056
  - `jvm.buffer.{memory.used,memory.limit,count}`
  - `jvm.system.cpu.{load_1m,utilization}` (used in runtime dashboard)
  - `jvm.memory.init`
- [ ] configure resource providers (deployment.environment copied to deployment.environment.name)
- [x] configure delta metrics temporality 
  - [x] https://github.com/elastic/elastic-otel-java/pull/1058 initial attempt with 'magic' behavior
  - [x] https://github.com/elastic/elastic-otel-java/pull/1062 set delta metrics temporality with explicit declarative config
- [x] configure span stacktrace capture #1067 
  - #1083 for improvements
- [x] configure default sampler #1067
- [x] configure indy enabled by default #1067
- [x] switch to alternative approach to avoid "magic" configuration: https://github.com/elastic/elastic-otel-java/pull/1060
- [x] universal profiling integration 
  - moved to a separate issue #1079 
- [x] inferred spans 
  - moved to a separate issue #1082 
- [x] doc/example: configure baggage to span attributes (opt-in)
- [x] doc/example: configure http/messaging span filtering (opt-in), this allows to fix #826 and #909 

### Open questions/ideas
- [ ] we can now easily print resulting config in json/yaml, should we use that instead for debug logging ?
  - might be easier to use than what we currently have (which also include all the defaults, whereas here we should have what has been explicitly set + what is modified/added by EDOT).
  - if we add this upstream it would allow an easy diff of the effective differences in behavior between the two implementations: for example if we replace a sampler, all the extensions that we explicitly configure, ...
[...]

PR #1056, opened 2026-04-17
decl. config. enable experimental runtime metrics

SylvainJuge:
Part of #1054 

Equivalent to add the following to declarative configuration
```yaml
instrumentation/development:
  java:
    runtime_telemetry:
      emit_experimental_metrics/development: true
```

Users can opt-out by setting it to `false` which will revert to the upstream default behavior with:
```yaml
instrumentation/development:
  java:
    runtime_telemetry:
      emit_experimental_metrics/development: false
```


custom/src/test/java/co/elastic/otel/declarativeconfig/ElasticDeclarativeConfigurationCustomizerTest.java
```diff
@@ -80,6 +83,12 @@ void defaultConfig() {
     assertThat(model.getTracerProvider()).isNull();
     assertThat(model.getMeterProvider()).isNull();
     assertThat(model.getLoggerProvider()).isNull();
+
+    // java experimental runtime metrics enabled by default
```
LikeTheSalad, line 87: I'm curious why this is enabled by default?

SylvainJuge: This is because some of the charts we include in dashboards rely on metrics that are not yet stable ([here in semconv](https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/#development)).

LikeTheSalad (approved):

PR #1060, opened 2026-04-20
switch alternative approach to declarative config

SylvainJuge:
part of #1054 

- removing the "magic" modification of runtime declarative configuration
  - keep the user-agent modification for now
  - keep the distribution attribute provider added transparently as it does not make sense to expose this to user-visible configuration and might create issues if not properly defined/used.
- provide a static yaml configuration included in the agent jar
- provide simple testing of yaml file
- add `--default-config-yaml` CLI option to access the embedded yaml configuration in agent jar.

possible follow-up:
- [x] add integration test to ensure the configuration is 100% correct and can be loaded by the agent, for example we can easily add a missing resource attribute provider without any failure with this implementation.

### DIFF
--- a/agent/entrypoint/src/main/java/co/elastic/otel/agent/ElasticAgent.java
+++ b/agent/entrypoint/src/main/java/co/elastic/otel/agent/ElasticAgent.java
@@ -19,7 +19,10 @@
 package co.elastic.otel.agent;
 
 import io.opentelemetry.javaagent.OpenTelemetryAgent;
+import java.io.IOException;
+import java.io.InputStream;
 import java.lang.instrument.Instrumentation;
+import java.util.Objects;
 
 /** Elastic agent entry point, delegates to OpenTelemetry agent */
 public class ElasticAgent {
@@ -56,7 +59,36 @@ public class ElasticAgent {
    * @param args arguments
    */
   public static void main(String[] args) {
-    OpenTelemetryAgent.main(args);
+    boolean upstreamDefault = true;
+    for (String arg : args) {
+      switch (arg) {
+        case "--default-config-yaml":
+          printDefaultConfigYaml();
+          upstreamDefault = false;
+          break;
+      }
+    }
+
+    if (upstreamDefault) {
+      OpenTelemetryAgent.main(args);
+    }
+  }
+
+  private static void printDefaultConfigYaml() {
+    try (InputStream input =
+        ElasticAgent.class
+            .getClassLoader()
+            .getResourceAsStream("inst/co/elastic/otel/config.yaml")) {
+      Objects.requireNonNull(input, "Default config yaml resource is missing");
+
+      byte[] buffer = new byte[8192];
+      int read;
+      while ((read = input.read(buffer)) != -1) {
+        System.out.write(buffer, 0, read);
+      }
+    } catch (IOException e) {
+      throw new IllegalStateException("Failed to print default config yaml", e);
+    }
   }
 
   private static void initLogging() {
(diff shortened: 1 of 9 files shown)


LikeTheSalad (dismissed):

agent/entrypoint/src/main/java/co/elastic/otel/agent/ElasticAgent.java
```diff
@@ -56,7 +59,36 @@ public static void agentmain(String agentArgs, Instrumentation inst) {
...
+      OpenTelemetryAgent.main(args);
+    }
+  }
+
+  private static void printDefaultConfigYaml() {
+    try (InputStream input =
+        ElasticAgent.class
+            .getClassLoader()
+            .getResourceAsStream("inst/co/elastic/otel/config.yaml")) {
+      Objects.requireNonNull(input, "Default config yaml resource is missing");
```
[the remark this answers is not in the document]
SylvainJuge: this is fine because the command is using the packaged path, which includes `/inst/' folder.

custom/src/main/java/co/elastic/otel/declarativeconfig/ElasticDeclarativeConfigurationCustomizer.java
```diff
@@ -61,49 +58,20 @@
...
 
-  private static final String RUNTIME_TELEMETRY = "runtime_telemetry";
-  private static final String EMIT_EXPERIMENTAL_METRICS_DEVELOPMENT =
-      "emit_experimental_metrics/development";
-
   @Override
   public void customize(DeclarativeConfigurationCustomizer customizer) {
     customizer.addModelCustomizer(
         model -> {
+          if (model == null) {
```
SylvainJuge, line 65: [for reviewer] this just prevents a confusing NPE when the configuration can't be parsed, hence making it harder to debug for the end-user.

custom/src/main/java/co/elastic/otel/declarativeconfig/ElasticDeclarativeConfigurationCustomizer.java
```diff
@@ -61,49 +58,20 @@
...
+            model = new OpenTelemetryConfigurationModel();
+          }
+
           customizeResources(model);
           customizeUserAgent(model);
-          customizeExperimentalRuntimeTelemetryMetrics(model);
```
SylvainJuge, line 74: [for reviewer] we remove the "magic" implementation, this is replaced with the default config in yaml (unfortunately we don't have a test for it, at least yet).

custom/src/test/java/co/elastic/otel/declarativeconfig/DefaultDeclarativeConfigTest.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.function.Consumer;
+import net.javacrumbs.jsonunit.assertj.JsonListAssert;
+import org.junit.jupiter.api.Test;
+
+public class DefaultDeclarativeConfigTest {
+
+  // For now, we can't override env variable for testing, thus we just verify the default values
+  // in the configuration we provide.
+  @Test
+  void testDefaults() {
```
SylvainJuge, line 39: [for reviewer] this tests that the yaml has a correct structure and contains the things we hope it contains, it does not really checks that everything in it is actually correct, which is what the smoke test provides.

custom/src/test/java/co/elastic/otel/declarativeconfig/ElasticDeclarativeConfigurationCustomizerTest.java
```diff
@@ -245,28 +236,6 @@ void addUserAgentPriority() {
         .contains("user-agent", "custom-user-Agent");
   }
 
-  @Test
-  void optOutExperimentalRuntimeMetrics() {
```
SylvainJuge, line 249: [for reviewer] I'm happy to get rid of this "half hadooken code"

LikeTheSalad (approved):

PR #1058, opened 2026-04-20
decl. config: delta metrics temporality

SylvainJuge:
part of #1054 

This makes EDOT java set `temporality_preference` to `delta` when not explicitly set by user
```yaml
meter_provider:
  readers:
    - periodic:
        exporter:
          # or 'otlp_grpc' for grpc
          otlp_http:
            endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318}/v1/metrics
            temporality_preference: delta
```

users can opt-out by setting `temporality_preference` explicitly:
```yaml
meter_provider:
  readers:
    - periodic:
        exporter:
          # or 'otlp_grpc' for grpc
          otlp_http:
            endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT:-http://localhost:4318}/v1/metrics
            # or 'low_memory'
            temporality_preference: cumulative
```

PR #1062, opened 2026-04-21
declarative config metrics delta

SylvainJuge:
part of #1054 

replaces #1058


smoke-tests/src/test/java/com/example/javaagent/smoketest/AppServerProvidersTest.java
```diff
@@ -60,15 +60,17 @@ static void afterAll() {
   @Test
   void tomcat() {
     target =
```
SylvainJuge, line 62: [for reviewer] this (and below) a test refactor to use builder pattern to make test slightly more readable/maintainable.

jackshirazi (approved):

PR #1064, opened 2026-04-23
move logger init to avoid ClassCircularityError

SylvainJuge:
Following upstream update to 2.27.0 we get some `ClassCircularityError` when the agent starts with java 17 (and surprisingly not with other JDK versions).

This PR moves the logger initialization to a latter stage to avoid the issue.

It's technically a bug, but hasn't been released yet.


jackshirazi:
again!

jackshirazi (approved):

PR #1067, opened 2026-04-23
declarative config: indy, propagator, stack trace, sampler

SylvainJuge:
Part of #1054 

- enable indy by default
- add default propagators
- configure span stacktrace
- configure sampler: parent-based consistent sampler (should be equivalent to the one we use from contrib).


SylvainJuge:
@coderabbitai review

SylvainJuge:
@coderabbitai review

jackshirazi (approved):

PR #1073, opened 2026-04-27
[ci] replace app-id to client-id to get token

SylvainJuge:
see warnings in https://github.com/elastic/elastic-otel-java/actions/runs/25001005570

For robots team, this might require a broader investigation given the amount of usage found: https://github.com/search?q=org%3Aelastic+language%3Ayaml+%22app-id%3A%22&type=code

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -19,7 +19,7 @@ jobs:
       id: get_token
       uses: actions/create-github-app-token@v3
       with:
-        app-id: ${{ secrets.OBS_AUTOMATION_APP_ID }}
+        client-id: ${{ secrets.OBS_AUTOMATION_APP_ID }}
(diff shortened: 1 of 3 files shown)


jackshirazi (approved):

v1v (approved):
We aim to use the alternative for `actions/create-github-app-token@v3`, since that's the former approach and we have been asked for using https://github.com/elastic/ci-gh-actions/tree/main/fetch-github-token.

Although, for simplicity, we have created https://github.com/elastic/oblt-actions/tree/main/github/create-token

SylvainJuge:
@v1v le me know if you have broader plans to migrate this, I only noticed the depreciation by accident when doing a release so this could have been silently ignored.

v1v:
> I only noticed the depreciation by accident when doing a release so this could have been silently ignored.

All good here, so I missed to say thanks for keeping those warnings under control 🙇 

We are planning the work in our end for the migration

PR #1089, opened 2026-04-30
provide experimental base declarative config support

### LINKED ISSUE #826: selectively ignore HTTP spans + configure it with declarative configuration

### LINKED ISSUE #909: Selectively ignore message queues
Users should be able to filter out specific messaging queues/topics from being traced. A list should provide one or more queue/topic names to ignore.

This option should support the wildcard *, which matches zero or more characters. Examples: /foo/*/bar/*/baz\*, *foo*. Matching is case insensitive by default. Prepending an element with (?-i) makes the matching case sensitive.

### LINKED ISSUE #1054: declarative config : base configuration of EDOT

SylvainJuge:
Closes #1054 by providing a first version of declarative configuration

Fixes #826
Fixes #909

- disable user-agent modification when using declarative configuration
- baggage to span attribute copy (example)
- filter http and messaging spans using rule-based sampler (examples)
- confiture span stacktrace capture extension (contrib)
- configure inferred spans (contrib), opt-in, disabled by default


custom/src/main/java/co/elastic/otel/declarativeconfig/ElasticDeclarativeConfigurationCustomizer.java
```diff
@@ -67,7 +67,6 @@ public void customize(DeclarativeConfigurationCustomizer customizer) {
           }
 
           customizeResources(model);
-          customizeUserAgent(model);
           return model;
```
jackshirazi, line 70: I don't remember this was the decision, that we'd remove the user-agent. I though it was absent from the declarative config but otherwise present?

SylvainJuge: From what I recall, we should "avoid magic and surprising users with things that are not explicit in declarative configuration".

This removes magically modifying the user-agent (if not set) in the declarative configuration, which was in fact the only option because if it's explicitly set in the declarative configuration then the version part of it would have to be maintained by end-user and thus likely to be error-prone.

For the automatic configuration though, we can leave it as-is, so it would not impact the current telemetry. A further improvement would be to then revise this telemetry to only rely on `telemetry.*` attributes within the data, and not on the `user-agent` header.

jackshirazi (approved):

PR #1091, opened 2026-05-01
Migrate sampler to non-contrib SDK incubator implementation

anuraaga:


jackshirazi:
there's no need for this. we'll migrate the entire dynamic control to the contributed implementation when that's available, then this sampler gets removed, so there's no point in doing this in the interim

anuraaga:
@jackshirazi - Just to confirm what do you mean by that? This is the implementation I wrote based on the spec vs contrib which was a pre-spec POC. Isn't it good to switch to the maintained implementation?

I mean I can send a PR to remove the contrib implementation which I think would be merged - would it help? 😉

jackshirazi:
This agent will move to use the dynamic control extension currently being added to contrib. That dynamic control extension already uses the SDK incubator sampler. So applying this PR would soon be pointless, so we'll avoid that additional work for now. Thanks for the contribution though, it would be right if we weren't aiming to switch relatively soon

PR #1094, opened 2026-05-04
add jdk 26 test

SylvainJuge:
- rename jdk test to `jdk-test` to easily identify JDK tests and make it easier to add them as branch protection rules required checks.
- addk JDK 26 to the test matrix

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -125,10 +125,15 @@ jobs:
 
   jdk-test:
     runs-on: ubuntu-latest
+    if: always()
     needs:
       - jdk-matrix
     steps:
-      - run: echo "no test to run, this job depends on the results of jdk-matrix"
+      - name: Check jdk-matrix result
+        run: |
+          result="${{ needs.jdk-matrix.result }}"
+          echo "jdk-matrix result: $result"
+          [[ "$result" == "success" ]] || exit 1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -119,6 +123,13 @@ jobs:
           name: test-results-${{ matrix.version }}-${{ matrix.vm }}
           path: '**/build/test-results/test/TEST-*.xml'
 
+  jdk-test:
+    runs-on: ubuntu-latest
+    needs:
```
v1v, line 129: @copilot , run always and fail if the matrix job jdk-test-matrix has any failures.

SylvainJuge: @v1v are you trying to make copilot write something like this ? 

```yaml
  jdk-test:
    runs-on: ubuntu-latest
    if: ${{ always() }}
    needs:
      - jdk-matrix
    steps:
      - run: |
          result="${{ needs.jdk-test-matrix.result }}"
          if [[ $result == "success" || $result == "skipped" ]]; then
            exit 0
          else
            exit 1
          fi
```

v1v: yep

v1v (approved):

LikeTheSalad (approved):

PR #1097, opened 2026-05-06
detect inferred spans on tracer name

SylvainJuge:
In https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2803 in contrib, we use the tracer name to filter the inferred spans.

In EDOT, we need to implement the same heuristic for consistency, also it allows to avoid relying on the internal, out of semconv and undocumented `is_inferred` span attribute.

Instead, we rely on the tracer name which is expected to be stable and is equal to `inferred-spans` for inferred spans.

This is mostly a refactor that anticipates that the `is_inferred` span attribute might change (or be removed) in the future, there is no visible change in behavior for end-users.

### DIFF
--- a/common/build.gradle.kts
+++ b/common/build.gradle.kts
@@ -7,16 +7,19 @@ plugins {
 description = "Elastic common utilities for OpenTelemetry Java"
 
 dependencies {
-    api("com.blogspot.mydailyjava:weak-lock-free:0.18")
-    compileOnly("io.opentelemetry:opentelemetry-sdk")
-    compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi")
-    implementation(libs.bundles.semconv)
+  api("com.blogspot.mydailyjava:weak-lock-free:0.18")
+  compileOnly("io.opentelemetry:opentelemetry-sdk")
+  compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi")
+  implementation(libs.bundles.semconv)
 
-    testImplementation("io.opentelemetry:opentelemetry-sdk")
-    testImplementation("io.opentelemetry:opentelemetry-sdk-testing")
-    testImplementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
-    testImplementation(project(":testing-common"))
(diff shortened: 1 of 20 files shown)


common/build.gradle.kts
```diff
@@ -7,16 +7,19 @@ plugins {
...
+  testImplementation("io.opentelemetry:opentelemetry-sdk")
+  testImplementation("io.opentelemetry:opentelemetry-sdk-testing")
+  testImplementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
+  testImplementation(project(":testing-common"))
+  testImplementation("io.opentelemetry:opentelemetry-exporter-otlp")
+  testImplementation("io.opentelemetry:opentelemetry-exporter-logging")
+  testImplementation("io.opentelemetry:opentelemetry-api-incubator")
+
+  // only to get reference to upstream inferred tracer name
+  testCompileOnly(libs.contribInferredSpans);
```
SylvainJuge, line 24: for reviewer: only this line is added, other changes are just reformatting.

common/src/main/java/co/elastic/otel/common/InferredSpanDetector.java
```diff
@@ -0,0 +1,30 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package co.elastic.otel.common;
+
+import io.opentelemetry.sdk.trace.ReadableSpan;
+
+public class InferredSpanDetector {
```
jackshirazi: do we really need a separated class just for this static method?

jackshirazi (dismissed):
approved, but note comment

common/src/main/java/co/elastic/otel/common/InferredSpanDetector.java
SylvainJuge: I haven't found any other utility class with static methods where to put this. Do you think it would be better to rename the class to `SpanUtil` to suggest that it may be extended with other span-related code in the future ?

jackshirazi (approved):

PR #1174, opened 2026-07-15
support cloudfoundry

jackshirazi:
This adds the maven downloadable path with each release to `cloudfoundry/index.yml`. This doesn't directly support cloudfoundry buildpacks, those need to be added to the cloudfoundry/java-buildpack repo, but when added that will use the `cloudfoundry/index.yml` file to provide agent downloads

When the cloudfoundry part of the buildpack support has been added, a separate doc update with changelog entry needs to be added, currently this has no changelog because it doesn't really add any feature as a standalone update


LikeTheSalad (dismissed):

.ci/release/post-release.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
v1v, line 1:
```bash
$ ci/release/post-release.sh
zsh: permission denied: .ci/release/post-release.sh
$ ls -l .ci/release/post-release.sh
-rw-r--r--  1 vmartinez  staff  923 15 Jul 15:26 .ci/release/post-release.sh
```

wonder if permissions should be changed

jackshirazi: excellent catch, thank you

v1v (approved):

LikeTheSalad (approved):

PR #1193, opened 2026-08-11
[release] remove deprecated changelog resolve

SylvainJuge:
Should fix warnings at release time : https://github.com/elastic/elastic-otel-java/actions/runs/30613292001

As suggested by @cotti


LikeTheSalad (approved):

LikeTheSalad:
[CI idea proposal] Maybe we should add some logic in the PR checks to avoid running the test suite when only markdown files are modified.

PR #1196, opened 2026-08-12
Skipping build for docs-only changes

LikeTheSalad:
As proposed [here](https://github.com/elastic/elastic-otel-java/pull/1193#issuecomment-5252541799).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,11 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - "**/*.md"
   pull_request:
+    paths-ignore:
+      - "**/*.md"
   workflow_dispatch:
 
 permissions:


SylvainJuge (approved):
It should be a good start, however I think it's probable branch protection rules and required checks might make those "docs only" PRs un-mergeable, so maybe check with the robots team first to see if there is a known reliable way to implement this.

LikeTheSalad:
Can someone from @elastic/observablt-ci confirm if these changes could cause issues down the road? Or if there's a better approach to achieve the same goal?

mallendem (approved):

v1v (commented):
IIUC, then the CI checks in the Branch/Rule Protections won't be honoured and PRs will not be merged without the required check.

We tried in the past using a hack:
- https://github.com/elastic/apm-agent-java/blob/main/.github/workflows/main-docs.yml#L1-L10
- https://github.com/elastic/apm-agent-java/blob/9f20e534a9381ce366b830d937aa2465cf710520/.github/workflows/main.yml#L43-L52

But I recall there were some edge-cases

PR #1237, opened 2026-09-16
Migrate changelog workflows to release-notes

Mpdreamz:
## What

Replaces the three legacy shared changelog workflows with the unified `release-notes` workflow from `elastic/docs-actions`, and updates all four release-notes pages to use the new explicit CDN product name syntax.

### Workflows deleted
- `changelog-init.yml` — validate + bespoke upstream-update job
- `changelog-submit.yml` — triggered by `workflow_run` on `changelog-init`
- `changelog-upload.yml` — bundled on push to main
- `changelog-upstream-update.yml` + `.ci/changelog-upstream-update.sh` — bespoke auto-entry generator, replaced by the standard `release-notes-changelog-file.yml`

### Workflows added
- **`release-notes.yml`** — unified validate/submit/bundle via `elastic/docs-actions/.github/workflows/release-notes.yml@v1`; bundle fires on GitHub release publish (`bundle-on-release: true`) instead of push to main
- **`release-notes-comments.yml`** — upserts PR changelog comments after `release-notes` completes
- **`release-notes-changelog-file.yml`** — auto-generates and commits changelog entries for PRs missing one, via `elastic/docs-actions/.github/workflows/release-notes-changelog-file.yml@v1`; replaces the bespoke `changelog-upstream-update` job and script

### Docs
- `docs/docset.yml`: added `release_notes: [edot-java]` so docs-builder prefetches the CDN bundles at startup
- `docs/release-notes/*.md` (4 files): changed `:::{changelog} /releases/` → `:::{changelog} edot-java` (explicit product name, CDN-sourced; the `/`-prefixed local path form is deprecated)

## Note

The PR comment and changelog-file workflows use a `workflow_run` trigger listening for `release-notes`. This only activates once these workflows land on `main` — the first PR opened after merge will get full comment coverage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/release-notes-comments.yml
```diff
@@ -0,0 +1,18 @@
...
+  actions: read
+# The PR number is stable; do not cancel — two upstreams can complete close
+# together and cancelling mid-upsert risks leaving the wrong comment body.
+concurrency:
+  group: changelog-comments-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: false
+jobs:
+  comment:
+    if: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion != 'cancelled'
+    uses: elastic/docs-actions/.github/workflows/release-notes-comments.yml@v1
```
v1v, line 18:
```suggestion
    uses: elastic/docs-actions/.github/workflows/release-notes-comments.yml@v1
    permissions:
      pull-requests: write
      actions: read
```

.github/workflows/release-notes-comments.yml
```diff
@@ -0,0 +1,18 @@
+name: Changelog PR comments
+on:
+  workflow_run:
+    # Must match the `name:` field of the release-notes workflow in this repo.
+    workflows: ["release-notes"]
+    types: [completed]
+permissions:
+  pull-requests: write
+  actions: read
```
v1v, line 9:
```suggestion
permissions: {}
```

.github/workflows/release-notes-changelog-file.yml
```diff
@@ -0,0 +1,12 @@
...
+  workflow_run:
+    # Must match the `name:` field of the release-notes workflow in this repo.
+    workflows: ["release-notes"]
+    types: [completed]
+permissions:
+  pull-requests: write
+  actions: read
+jobs:
+  changelog-file:
+    uses: elastic/docs-actions/.github/workflows/release-notes-changelog-file.yml@v1
```
v1v, line 12:
```suggestion
    uses: elastic/docs-actions/.github/workflows/release-notes-changelog-file.yml@v1
    permissions:
      pull-requests: write
      actions: read
```

.github/workflows/release-notes-changelog-file.yml
```diff
@@ -0,0 +1,12 @@
+name: Changelog file
+on:
+  workflow_run:
+    # Must match the `name:` field of the release-notes workflow in this repo.
+    workflows: ["release-notes"]
+    types: [completed]
+permissions:
+  pull-requests: write
+  actions: read
```
v1v, line 9:
```suggestion
permissions: {}
```

v1v (commented):
some nits, to use the same permission strategy where top-level is empty and the ones in the jobs are defined with more granularity

theletterf:
Hey, I was looking at this and had a couple of questions (happy to be wrong on either).

**On the bundle strategy**: `release-step-3.yml` creates the GitHub release with `GH_TOKEN: ${{ github.token }}`. GitHub doesn't fire `on: release` for events created by `GITHUB_TOKEN`, so `bundle-on-release: true` in `release-notes.yml` would never trigger. Would it make sense to add the bundle jobs directly inside `release-step-3`, after the `create-github-release` step, instead?

**On the render pages**: the four `docs/release-notes/*.md` files are now pointing at `:::{changelog} edot-java` on the CDN, but I don't think any bundles have been published there yet. The old `changelog-upload.yml` uploaded entries, not release bundles. The `build / build` check seems to be failing because of a CDN 404 on that. Could those four files (and the `release_notes:` block in `docset.yml`) stay on the local path for now, and move to CDN in a follow-up once the first bundle is published?

.github/workflows/release-notes-comments.yml
Mpdreamz: Done — moved to job level on both files.

v1v (dismissed):

.github/workflows/release-notes-changelog-file.yml
```diff
@@ -0,0 +1,13 @@
+name: Changelog file
```
SylvainJuge, line 1: do you think we could simplify and only have one workflow for both "comment" or "pushing the changelog file" when relevant ? It's quite likely to be replicated across repositories so having a single top-level GH action would make this simpler and avoid duplication.

Here they have the same run triggers and the same permissions on the PR.

.github/workflows/release-step-3.yml
```diff
@@ -262,3 +262,43 @@ jobs:
...
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ inputs.ref }}
+          persist-credentials: false
+      - name: Create bundle
+        id: create
+        uses: elastic/docs-actions/changelog/bundle-create-version@v1
+        with:
+          version: ${{ env.RELEASE_VERSION_TAG }}
+          github-token: ${{ github.token }}
```
SylvainJuge, line 286: To implement https://github.com/elastic/elastic-otel-java/issues/1238 would it be just a matter of modifying the bundle file just between those two steps ? Or is it something we would need to do on a separate workflow ?

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,14 @@
...
+  push:
+    branches: [main]
+permissions: {}
+jobs:
+  release-notes:
+    permissions:
+      contents: read
+      pull-requests: read
+      packages: read
+    uses: elastic/docs-actions/.github/workflows/release-notes.yml@v1
```
SylvainJuge, line 14: Is it only checking the PR has a changelog file and/or the appropriate labels are applied here ?

Mpdreamz:
**Re: consolidating the two downstream workflows** — done, merged `release-notes-comments.yml` and `release-notes-changelog-file.yml` into a single `release-notes-downstream.yml` with two jobs. The `comment` job keeps its `if:` guard (only fires on PR events). Pushed in the latest commit.

**Re: #1238 and the bundle description** — not straightforward between those two steps. The old `pre-post-release.yml` bundle step ran `.ci/changelog-bundle-description.sh` to inject upstream dependency versions into the bundle's description field; that's exactly what's now missing. `bundle-create-version@v1` drives the `docs-builder changelog gh-release` command, which doesn't yet accept a custom description. When docs-builder adds that input, the bundle job in `release-step-3.yml` is the right place to wire it in — either as an additional step before `Create bundle`, or as a new action input. Tracked in #1238.

**Re: what `release-notes.yml` does** — three things: (1) on a PR event it validates the PR has either a changelog entry file or a `changelog:skip` label; (2) on push to main it uploads the merged PR's changelog entry to the CDN so it's available for the release-time bundle; (3) it does nothing else — bundle creation happens separately in `release-step-3.yml` after the GitHub release is cut.

Also fixed a related issue spotted in review: `pre-post-release.yml` was still running `docs-builder changelog bundle` in the pre-release phase, writing a `docs/releases/<version>.yaml` file into the repo. That file was referenced by the old release notes markdown. Now that release notes use CDN sourcing (`:::{changelog} edot-java`), the committed file was dead code — removed that step and the `docs-builder/setup@v1` that only served it.

.github/workflows/release-notes-changelog-file.yml
Mpdreamz: Keeping them separate is intentional — the onboarding guide documents these as two distinct files to copy, and repos can opt into either one independently (comments without auto-commit, or vice versa). I merged them in a previous commit and then reverted it; sorry for the noise, that was Claude being too eager.

.github/workflows/release-step-3.yml
Mpdreamz: The plan is to add a `description` input to `elastic/docs-actions/changelog/bundle-create-version@v1` itself, then wire the output of `.ci/changelog-bundle-description.sh` into it here in `release-step-3.yml`. That keeps the description logic in this repo and avoids rebuilding it inside the action. Tracked in #1238.

.github/workflows/release-notes.yml
Mpdreamz: On a PR event it always validates labels, checks whether a changelog file is present (if the labels *and* workflow config require one), and if a file is present validates its name format and contents. On push to main it uploads the merged PR's changelog entry to the CDN. Bundle creation is separate, handled by `bundle-create-version@v1` in `release-step-3.yml` after the GitHub release is cut.

.github/workflows/release-notes-changelog-file.yml
Mpdreamz: Also both are completely optional if you want neither comments or automated PR checkins, this limits blast radius a bit for repos that just want PR checks bundle automation.

.github/workflows/release-step-3.yml
Mpdreamz: Done — elastic/docs-actions#367 adds a `description` input to `bundle-create-version@v1`, and this PR now runs `.ci/changelog-bundle-description.sh` before `Create bundle` and passes its output as that input. Ties off the last item from the old pre-release bundle step.

.github/workflows/release-notes-changelog-file.yml
SylvainJuge: Fine for me to have both, it's just a bit confusing when reviewing the PR.

.github/workflows/release-notes.yml
SylvainJuge: Thanks for the explanation, that's more clear.

.github/workflows/release-step-3.yml
SylvainJuge: Thanks ! This should allow to close #1238 once we've made sure it works as expected with the next release.

SylvainJuge (approved):

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,14 @@
+name: release-notes
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited, labeled, unlabeled]
+  push:
+    branches: [main]
+permissions: {}
+jobs:
+  release-notes:
+    permissions:
```
LikeTheSalad: Don't we need to include `id-token: write` here for the upstream "Authenticate with AWS" step?

.github/workflows/release-step-3.yml
```diff
@@ -262,3 +262,47 @@ jobs:
...
+    outputs:
+      bundle-path: ${{ steps.create.outputs.bundle-path }}
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ inputs.ref }}
+          persist-credentials: false
+      - name: Generate bundle description
+        id: description
+        run: echo "text=$(.ci/changelog-bundle-description.sh)" >> "$GITHUB_OUTPUT"
```
LikeTheSalad: Does this script return a text with newlines? If so, wouldn't it cause parsing issues later for $GITHUB_OUTPUT?

SylvainJuge: I can confirm there are newlines in the output as it's few lines of markdown.

